PR #159, opened 2013-06-25
Initial test for interactive in_process

jarl-dk:
This test illustrates that interaction does not work with `@in-process` tag.


dg-ratiodata:
I think this never can work, since ruby is blocked by reading from stdin. But writing to stdin will not be reached because ruby is blocked.

PR #161, opened 2013-07-04
Reporting argument error

cpb:
Given I have pygments, bcat, and rdiscount installed
When I ran my cucumber features with ARUBA_REPORT_DIR=doc
Then I received an ArgumentError from inside lib/aruba/reporting.rb

I poked around the history of ```SpawnProcess``` and ```Reporting``` to try and confirm that you dropped the ```filter_ansi``` flag support, seems so.

I suppose this'll take more work to make Travis happy... I tried tagging it as ```@wip``` for jruby, but i didn't take a close look at how to do that right.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ rvm:
 matrix:
   allow_failures:
     - rvm: rbx
+
+before_install: "sudo easy_install Pygments"
+
 notifications:
   email:
     - cukes-devs@googlegroups.com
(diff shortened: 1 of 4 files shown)


dg-ratiodata:
Is this issue still valid? @jarl-dk Close?

maxmeyer:
If we decide to implement reporting as a formatter for cucumber - see #256 - this PR will be obsolete. I also prefer to get rid of pygmentize and use https://github.com/jneen/rouge instead which is pure ruby.

dg-ratiodata:
Aruba/Reporting is deprecated and will be removed.

PR #214, opened 2014-09-17
Add file md5 check support

ipmsteven:


dg-ratiodata:
@jarl-dk 

I see no reason to change the current implementation for using md5 or some other hash algorithm. I think `FileUtils.compare_file` is readable/sufficient enough and does not pull in other dependencies like `digest/sha2` or so.

Maybe it would be worth to setup somelike https://github.com/middleman/middleman-guides based on `middleman` or `jekyll` at [github-pages](https://pages.github.com/). This would lower the need to merge each and every piece of code but also empower users of `aruba` to do some more advanced stuff using this documentation.

Would do you think @jarl-dk? @ipmsteven would this helpful for you?

jarl-dk:
@dg-ratiodata : As I understand your latest comment

  1. You are satisified with the new file comparrison feature and take back your original feature request for MD5 check. Is that correctly understood?
  1. Not related to MD5 checking; you propose a new approach for a more community-oriented documentation approach. Is that correct?

Please confirm.

dg-ratiodata:
1. Correct
2. Correct -> To reduce the need for more and more application specific method

jarl-dk:
I close this PR.

You are encourage to follow up (2) in another PR.

PR #217, opened 2014-09-25
Use simplecov for coverage reports for specs and cucumber

limhoff-r7:
simplecov is a wrapper on the built-in coverage library for Ruby 1.9+.  When running `rake spec` or `rake cucumber`, a coverage report will be output in coverage/index.html.  If you run both tasks or all tasks with `rake spec` within a close enough time window (I think 10 minutes) the results will all be merged so you can get coverage across both the specs and features.  If this PR is accepts, then I'll add another to use coveralls.io to publish the coverage reports to coveralls.io, but someone with admin access will have to setup the hooks on coveralls.io for that.  Additionally, the reports from simplecov are a confirmation that not all steps are being exercised, include the ones dealing with rvm.  I have a fix for that too that I can post in a later PR.

screen shot 2014-09-25 at 2 04 16 pm

# Verifications Steps

## specs

- [ ] `rake spec`
- [ ] VERIFY no failures
- [ ] `open coverage/index.html`
- [ ] VERIFY coverage report includes specs
- [ ] `rm -rf coverage`

## cucumber

- [ ] `rake cucumber`
- [ ] VERIFY no failures
- [ ] `open coverage/index.html`
- [ ] VIERFY coverage report includes features and steps
- [ ] `rm -rf coverage`

## all

- [ ] `rake`
- [ ] VERIFY no failures
- [ ] `open coverage/index.html`
- [ ] VERIFY coverage includes specs, features, and steps

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,8 +1,40 @@
-SimpleCov.start do
-  add_filter "/features/"
-  add_filter "/spec/"
-  add_filter "/tmp"
-  add_filter "/vendor"
-  
-  add_group "lib", "lib"
+# RM_INFO is set when using Rubymine.  In Rubymine, starting SimpleCov is
+# controlled by running with coverage, so don't explicitly start coverage (and
+# therefore generate a report) when in Rubymine.  This _will_ generate a report
+# whenever `rake spec` is run.
+SimpleCov.start unless ENV.key? 'ARUBA_NO_COVERAGE'
+
+SimpleCov.configure do
+  # ignore this file
+  add_filter '.simplecov'
+
+  # Rake tasks aren't tested with rspec
+  add_filter 'Rakefile'
+  add_filter 'lib/tasks'
+
+  #
+  # Changed Files in Git Group
+  # @see http://fredwu.me/post/35625566267/simplecov-test-coverage-for-changed-files-only
+  #
+
+  untracked = `git ls-files --exclude-standard --others`
+  unstaged = `git diff --name-only`
+  staged = `git diff --name-only --cached`
+  all = untracked + unstaged + staged
+  changed_filenames = all.split("\n")
+
+  add_group 'Changed' do |source_file|
+    changed_filenames.select do |changed_filename|
+      source_file.filename.end_with?(changed_filename)
+    end
+  end
+
+  add_group 'Libraries', 'lib'
+
+  #
+  # Specs are reported on to ensure that all examples are being run and all
+  # lets, befores, afters, etc are being used.
+  #
+
+  add_group 'Specs', 'spec'
 end
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,4 +1,35 @@
 $LOAD_PATH.unshift(File.dirname(__FILE__) + '/../../lib')
-require 'aruba/cucumber'
+
+# Has to be the first file required so that all other files show coverage information
+require 'simplecov'
+
+#
+# Standard Library
+#
+
 require 'fileutils'
+require 'pathname'
+
+#
+# Gems
+#
+
+require 'aruba/cucumber'
 require 'rspec/expectations'
+
+Before do |scenario|
+  command_name = case scenario
+                 when Cucumber::RunningTestCase::Scenario
+                   "#{scenario.feature.title} #{scenario.name}"
+                 else
+                   raise TypeError.new("Don't know how to extract command name from #{scenario.class}")
+                 end
+
+  # Used in simplecov_setup so that each scenario has a different name and their coverage results are merged instead
+  # of overwriting each other as 'Cucumber Features'
+  set_env('SIMPLECOV_COMMAND_NAME', command_name)
+
+  simplecov_setup_pathname = Pathname.new(__FILE__).expand_path.parent.join('simplecov_setup')
+  # set environment variable so child processes will merge their coverage data with parent process's coverage data.
+  set_env('RUBYOPT', "-r#{simplecov_setup_pathname} #{ENV['RUBYOPT']}")
+end
--- a/features/support/simplecov_setup.rb
+++ b/features/support/simplecov_setup.rb
@@ -1,0 +1,9 @@
+# @note this file is loaded in env.rb to setup simplecov using RUBYOPTs for child processes and @in-process
+
+require 'simplecov'
+
+root = File.expand_path('../../../', __FILE__)
+
+SimpleCov.command_name(ENV['SIMPLECOV_COMMAND_NAME'])
+SimpleCov.root(root)
+load File.join(root, '.simplecov')
--- a/spec/aruba/spawn_process_spec.rb
+++ b/spec/aruba/spawn_process_spec.rb
@@ -3,7 +3,7 @@ require 'aruba/spawn_process'
 module Aruba
   describe SpawnProcess do
 
-    let(:process) { SpawnProcess.new('echo "yo"', 0.1, 0.1) }
+    let(:process) { SpawnProcess.new('echo "yo"', 1, 1) }
 
     describe "#stdout" do
       before { process.run! }


.simplecov
```diff
@@ -0,0 +1,42 @@
+# RM_INFO is set when using Rubymine.  In Rubymine, starting SimpleCov is
+# controlled by running with coverage, so don't explicitly start coverage (and
+# therefore generate a report) when in Rubymine.  This _will_ generate a report
+# whenever `rake spec` is run.
+unless ENV['RM_INFO']
```
dg-ratiodata: make this a single line statement...

.simplecov
```diff
@@ -0,0 +1,42 @@
...
+  #
+
+  untracked = `git ls-files --exclude-standard --others`
+  unstaged = `git diff --name-only`
+  staged = `git diff --name-only --cached`
+  all = untracked + unstaged + staged
+  changed_filenames = all.split("\n")
+
+  add_group 'Changed' do |source_file|
+    changed_filenames.detect { |changed_filename|
```
dg-ratiodata: isn't `select` the preferred choice for this? -> rubocop offense?

features/support/env.rb
```diff
@@ -1,5 +1,35 @@
...
+# Standard Library
+#
+
 require 'fileutils'
+require 'pathname'
+
+require 'aruba/cucumber'
 require 'rspec/expectations'
 
+Before do |scenario|
```
dg-ratiodata: Can this be moved to `simplecov_setup`?

features/support/env.rb
```diff
@@ -1,5 +1,35 @@
 $LOAD_PATH.unshift(File.dirname(__FILE__) + '/../../lib')
-require 'aruba/cucumber'
+
+# Has to be the first file required so that all other files show coverage information
+require 'simplecov'
```
dg-ratiodata: Is this `require` required? It's already in `simplecov_setup`.

features/support/simplecov_setup.rb
```diff
@@ -0,0 +1,12 @@
+# @note this file is loaded in env.rb to setup simplecov using RUBYOPTs for child processes
+
+require 'simplecov'
+
+require 'pathname'
+
+root = Pathname(__FILE__).expand_path.parent.parent.parent
```
dg-ratiodata: I would prefer this here:

`File.expand_path('../../../', __FILE__)`

So you don't need to require 'pathname'.

features/support/simplecov_setup.rb
```diff
@@ -0,0 +1,12 @@
...
+
+SimpleCov.command_name(ENV['SIMPLECOV_COMMAND_NAME'])
+SimpleCov.root(root)
+load root.join('.simplecov')
```
dg-ratiodata: is `load` needed here? Does 'simplecov` not load it itself?

.gitignore
```diff
@@ -10,6 +10,9 @@ tmtags
 \#*
 .\#*
 
+## RUBYMINE
```
dg-ratiodata: This is RUBYMINE-specific what about all other development-environments? Can you build a solution on your own or is it needed to add it to `aruba`?

dg-ratiodata:
Can you rebase with the most recent `master`?

.gitignore
limhoff-r7: .idea directories are used by all Jetbrains IDEs.  I added it to the ignores to be on par with the VIM ignores on line 16.

.simplecov
limhoff-r7: I use vim and Rubymine for Ruby development.  I have no idea how other IDEs coverage tools work, so yes, it's possible, but RM_INFO an environment variable that is set when Rubymine runs, so it's what I use.  We can't pick an environment variable and have all the IDEs agree on it and have it work now.  (We can petition the IDEs to agree on something, but they're not doing it now.) RM_INFO is used because it is already defined by the IDE.  If you want support for other IDEs people will have to submit PRs for whatever environment variable their IDEs use when they run rake spec or rake cucumber from inside the IDE.

.simplecov
limhoff-r7: I'll make the change to `select`.  I use the [Jim Weirich](http://devblog.avdi.org/2011/07/26/the-procedurefunction-block-convention-in-ruby/) style for blocks: if a block has a yieldreturn, then I use `{ }` and if it's only run for side-effects, then I use `do end`.  If you're following the other style of `{ }` for one-liners (which I don't do anyway since you can't place a breakpoint on the inside of the block without breakpointing on the outside) and `do end` for multiline, then I'll change this to `do end` to conform to the project style.

features/support/env.rb
limhoff-r7: This is only necessary if we want coverage in the process the cucumber process and not the process under test, so it can be removed.  I think I made that change in the source project and didn't update this PR because the review process was stalled.  I only meant for `simplecov_setup` to be used in child process and not the cucumber process itself.

features/support/env.rb
limhoff-r7: I only meant to `simplecov_setup` to be loaded in the child process, hence why it's path is used in RUBYOPT on line 34.  I don't want it loaded in the main cucumber process.

features/support/simplecov_setup.rb
limhoff-r7: Ok

features/support/simplecov_setup.rb
limhoff-r7: This is loading the `.simplecov` configuration file.  Yes, it is required because when `require 'simplecov'` is called, Simplecov will try to load `'.simplecov'`, but the working directory will be wrong (because it will be tmp/aruba), so I have to explicitly set the root first, `SimpleCov.root(root)` and then load the configuration file `load root.join('.simplecov')`

features/support/env.rb
limhoff-r7: Is there a better place to put `simplecov_setup` so it's not loaded by the cucumber process, but available for require by child processes?  Should I make a different directory than `features/support`?

limhoff-r7:
Rebased

features/support/env.rb
limhoff-r7: Reviewing more of the support files and features, some of them are tagged `@in-process` and the only way to get in-process coverage in cucumber is for simplecov to be loaded in the cucumber process, so this require should stay for the @in-process tag.  If there wasn't that tag, it could have been removed.

.simplecov
```diff
@@ -0,0 +1,40 @@
...
+
+  add_group 'Libraries', 'lib'
+
+  #
+  # Specs are reported on to ensure that all examples are being run and all
+  # lets, befores, afters, etc are being used.
+  #
+
+  add_group 'Specs', 'spec'
+end
```
dg-ratiodata: Newline is missing at the end of the file.

KronicDeth: Added

dg-ratiodata:
@jarl-dk What do you think?

.simplecov
```diff
@@ -0,0 +1,40 @@
+# RM_INFO is set when using Rubymine.  In Rubymine, starting SimpleCov is
+# controlled by running with coverage, so don't explicitly start coverage (and
+# therefore generate a report) when in Rubymine.  This _will_ generate a report
+# whenever `rake spec` is run.
+SimpleCov.start unless ENV['RM_INFO']
```
maxmeyer: I'm still not happy introducing code which is only for a "single" developer's development environment - no offense :smile:.

```ruby
SimpleCov.start unless ENV.key? 'ARUBA_NO_COVERAGE'
```
would be ok for me. You can then setup a rake task which sets `ENV['ARUBA_NO_COVERAGE']` if `ENV['RM_INFO']` is set.

We would set the `ARUBA_COVERAGE` then by default in our rake test task.


@mattwynne @jarl-dk 

What do you think?

mattwynne: I agree.

maxmeyer:
@limhoff-r7 

If you like, please fix the things mentioned above + rebase again + squash commits - all changes which belong together should be in a single commit:

* Integrate into spec
* Integrate into features
* [...]

And so on. All "helper" commits like "Fixed offense" should be gone after this. Then I'll be happy to merge this.

Thanks a lot.

.simplecov
jarl-dk: I agree.

limhoff-r7:
Rebase and squash complete.

limhoff-r7:
The squash lost the ENV variable rename.  I'll amend and force push again.

limhoff-r7:
OK, environment variable is correct now.

maxmeyer:
Merged. Thanks a lot.

PR #218, opened 2014-10-09
Accept single and multiple files per method invocation

dg-ratiodata:
There were methods which accept only a single file per method invocation and there were methods which accept multiple files (as arguments). I added functionality which makes it possible to use multiple files per method invocation. One does not need to put those file names in an array. I use the splat operator to make this possible. Though it is still possible to use an array of file names, because I "flatten" the file names array. The only "drawback" is, that it is not possible to set default values anymore if the splat operator is in use. I work around this by checking via `#nil?`.

```ruby
touch_file 'file1', 'file2'
```

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -88,20 +88,24 @@ module Aruba
     #
     # @param [String] file_content
     #   The content which should be written to the file
-    def write_file(file_name, file_content)
-      _create_file(file_name, file_content, false)
+    def write_file(*file_names, file_content)
+      file_names.flatten.each do |f|
+        _create_file(f, file_content, false)
+      end
     end
 
     # Create an empty file
     #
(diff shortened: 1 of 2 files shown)


dg-ratiodata:
The only thing would that the `aruba`-methods behave like there "real world"-pendants on the shell.

```bash
touch_file 'file1', 'file2' => touch file1 file2
remove_file 'file1', 'file2' => rm file1 file2
```

jarl-dk:
Good point. I also see that the `FileUtils` API supports this. What about forwarding all args to the equivalent FilUtils API method? That way aruba API will also support the `options` hash. Would you mind change it to exploit the `FileUtils` API that way?

dg-ratiodata:
No problem. :smile:

dg-ratiodata:
Outdated by #229

PR #219, opened 2014-10-16
Add copy method to aruba

dg-ratiodata:
In one of my projects I need to copy files around during testing. I added a method to do this.

@jarl-dk Would you mind to review this?


jarl-dk:
Taking the fixture expansion feature out of this PR discussion. I think this feature (copy) should work just like the other file operation features, that is paths are relative to current directory with home directory expansion using `~`.

dg-ratiodata:
@jarl-dk 

I will rework this when the other PRs are merged/declined. Does this make sense?

dg-ratiodata:
@jarl-dk @mattwynne 

I refactored the PR and would like to ask you for feedback. Each commit changes separate bits of aruba.

dg-ratiodata:
@mattwynne 
Any chance that you review this today, as @jarl-dk is busy ? If you're ok with this PR, I'm happy to merge it myself.

lib/aruba/api.rb
```diff
@@ -206,6 +206,46 @@ def touch(*args)
...
+    #   A file or directory name. If multiple sources are given the destination
+    #   needs to be a directory
+    #
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def copy(*source, destination)
+      source = source.flatten
+
+      source.each do |s|
+        raise ArgumentError, %(The following source "#{s}" does not exist.) unless exist? s
```
mattwynne: How about we extract this into `#assert_exists`?

lib/aruba/api.rb
```diff
@@ -206,6 +206,46 @@ def touch(*args)
...
+      end
+
+      raise ArgumentError, "Using a fixture as destination (#{destination}) is not supported" if destination.start_with? FIXTURES_PATH_PREFIX
```
mattwynne: Could this logic just go into `#expand_path`?

lib/aruba/api.rb
```diff
@@ -206,6 +206,46 @@ def touch(*args)
...
+      raise ArgumentError, "Multiples sources can only be copied to a directory" if source.count > 1 && exist?(destination) && !directory?(destination)
```
mattwynne: How does it behave if there's only one source but the destination is a file? Does it overwrite the file?

lib/aruba/api.rb
```diff
@@ -206,6 +206,46 @@ def touch(*args)
...
+      if source_paths.count > 1
+        _mkdir(destination_path)
+      else
+        _mkdir(File.dirname(destination_path))
+        source_paths = source_paths.first
+      end
+
+      FileUtils.cp_r source_paths, destination_path
+
+      self
```
mattwynne: :clap:

I love this pattern for command methods. Let's use it everywhere!

lib/aruba/api.rb
```diff
@@ -206,6 +206,46 @@ def touch(*args)
...
+    #   files and directories. If multiple sources are given the destination
+    #   needs to be a directory
+    #
+    # @param [String] destination
+    #   A file or directory name. If multiple sources are given the destination
+    #   needs to be a directory
+    #
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
+    def copy(*source, destination)
```
mattwynne: I have another suggestion on the API, which is that we use an API more like Capybara's:

`copy 'foo', 'bar', to: 'baz/'

WDYT?

lib/aruba/api.rb
maxmeyer: Mmmh... `assert_*` reads more like `Test::Unit` I prefer to leave it like that.

lib/aruba/api.rb
maxmeyer: On other methods we try to make them "look" like shell commands. I would like to leave it like that. WDYT?

lib/aruba/api.rb
maxmeyer: Good question. I asked this myself and decided to do it like `cp` which overwrites it.

lib/aruba/api.rb
maxmeyer: Yes. I tried that for each new "command"-method I added.

lib/aruba/api.rb
maxmeyer: Mmmh, I'm not sure how this can be possible. If  a string does not start with `%` it is a path which needs to be expanded. The check `raise ArgumentError, "Using a fixture as destination (#{destination}) is not supported" if destination.start_with? FIXTURES_PATH_PREFIX` is specific for `copy`.

maxmeyer:
I rebased this PR.

maxmeyer:
Can we go on with that?

mattwynne:
Go for it!

PR #220, opened 2014-10-26
Add force flag to remove_dir and remove_file

dg-ratiodata:
It was not possible to force `#remove_file` to "delete" an not existing file. It just raised an error. I added a `force`-option, so that no error is raised if a file/directory does not exists.

To make `#remove_dir` backward-compatible I set `force` to `true`, because `Dir.rmdir` does not care if a directory exists or not.

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -224,11 +224,15 @@ module Aruba
     #
     # @param [String] file_name
     #    The file which should be deleted in current directory
-    def remove_file(file_name)
+    def remove_file(file_name, force = false)
       in_current_dir do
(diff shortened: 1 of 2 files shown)


dg-ratiodata:
I thought about what you wrote. Instead of `rmdir` we can use `rm_r` which does the same, but using `rm_rf` the implementation would be the same for file and directory.

dg-ratiodata:
Cleaned up and rebased.

dg-ratiodata:
@jarl-dk Can we go on with this PR?l :smile_cat:

dg-ratiodata:
Outdated by #229

PR #224, opened 2015-01-15
Add fixtures to aruba

dg-ratiodata:
This PR adds the basics to support fixtures in `aruba`. Also see  #219.

Things to discuss:

* [ ] Use of `OpenStruct`
  
  I used `OpenStruct` here to make both paths of a fixture available: The relative one which is used to find a fixture and the absolute one, which can be used by a `consumer` of the api like `copy_file`. I can replace this by a `Hash` if this is preferred. Though `OpenStruct` is part of `ruby core distribution`.
  
* [ ] Use of `Pathname`

  I like the `relative_path_from`-functionality of `Pathname`. It makes the intention of the author clear and is quite easy to use to build relative paths.

* [ ] Use of `%/` as fixture path prefix

  It does not have a special meaning in ruby so far - besides percent classes which are not followed by `/`.

* [ ] Order of directories to be searched
  
  I use "my" preferred order here, but I'm open to change it.

* [ ] Commit Style
  
  I will refactor the "commits" after all things above are discussed to make sure that there's one commit for the tests and one commit for the implementation.

* [ ] Rubocop

  I think we will need to "guard" this [line](https://github.com/cucumber/aruba/pull/224/files#diff-65af44ccc0b346dd91178a015aee5560R39) and that [line](https://github.com/cucumber/aruba/pull/224/files#diff-65af44ccc0b346dd91178a015aee5560R786) here because of `===` (3x `=`). But `===` returns `true'/`false` for both `String` AND `Regexp` and thus is more flexible than `==`

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -3,6 +3,8 @@ require 'rbconfig'
 require 'rspec/expectations'
 require 'aruba'
 require 'aruba/config'
+require 'ostruct'
+require 'pathname'
 
 Dir.glob( File.join( File.expand_path( '../matchers' , __FILE__ )  , '*.rb' ) ).each { |rb| require rb }
 
@@ -21,20 +23,32 @@ module Aruba
     # @example Single file name
     #
     #   # => <path>/tmp/aruba/file
-    #   absolute_path('file')
+    #   expand_path('file')
     #
     # @example Single Dot
     #
     #   # => <path>/tmp/aruba
-    #   absolute_path('.')
+    #   expand_path('.')
     #
     # @example Join and Expand path
     #
     #   # => <path>/tmp/aruba/path/file
-    #   absolute_path('path', 'file')
+    #   expand_path('path', 'file')
     #
+    def expand_path(*args)
+      if %r{^#{fixtures_path_prefix}/} === args.first
+        File.join fixtures_directory, args.shift.sub(%r{^#{fixtures_path_prefix}/}, ''), *args
+      else
+        in_current_directory { File.expand_path File.join(*args) }
+      end
+    end
+
+    # @private
+    # @deprecated
     def absolute_path(*args)
-      in_current_directory { File.expand_path File.join(*args) }
+      warn('The use of "absolute_path" is deprecated. Use "expand_path" instead')
+
+      expand_path(*args)
     end
 
     # Execute block in current directory
@@ -126,7 +140,7 @@ module Aruba
                   {}
                 end
 
-      args = args.map { |p| absolute_path(p) }
+      args = args.map { |p| expand_path(p) }
       args.each { |p| _mkdir(File.dirname(p)) }
 
       FileUtils.touch(args, options)
@@ -194,7 +208,7 @@ module Aruba
                mode
              end
 
-      args = args.map { |p| absolute_path(p) }
+      args = args.map { |p| expand_path(p) }
       args.each { |p| raise "Expected #{p} to be present" unless FileTest.exists?(p) }
 
       FileUtils.chmod(mode, args, options)
@@ -224,7 +238,7 @@ module Aruba
       expected_permissions = args.shift
       expected_result      = args.pop
 
-      args = args.map { |p| absolute_path(p) }
+      args = args.map { |p| expand_path(p) }
 
       args.each do |p|
         raise "Expected #{p} to be present" unless FileTest.exists?(p)
@@ -269,7 +283,7 @@ module Aruba
                   {}
                 end
 
-      args = args.map { |p| absolute_path(p) }
+      args = args.map { |p| expand_path(p) }
 
       FileUtils.rm(args, options)
     end
@@ -322,7 +336,7 @@ module Aruba
                   {}
                 end
 
-      args = args.map { |p| absolute_path(p) }
+      args = args.map { |p| expand_path(p) }
 
       FileUtils.rm_r(args, options)
     end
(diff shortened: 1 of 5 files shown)


dg-ratiodata:
Any comments about this one @jarl-dk ?

dg-ratiodata:
@jarl-dk ping... :smile:

maxmeyer:
This PR has been opened for a while. 

@aslakhellesoy @mattwynne @jarl-dk Is there anyone willing to review this? I'm a bit frustrated waiting so long for feedback.

mattwynne:
I don't understand what a fixture is in this context. Is there an example somewhere of how this feature would be used in practice? Maybe add something to the README?

dg-ratiodata:
This PR is a result of the discussion in #219.

I added an example to the readme. Though I must admit, that this PR only adds some more basic stuff which will become really useful when #219 (Add copy method) is finished and makes use of this. Until then the example in the readme feels a little bit unfinished.

jarl-dk:
I'll take a look soon.

jarl-dk:
@mattwynne : In this context a fixture is a constant byte array (stored as a file). Such fixture files are sometimes used to compare binary data against. It could be images or other binary static data that you verify against in a test. e.g.
`check_binary_file_content('produced_file.jpeg','fixtures/expected_file_content.jpeg')`
such fixtures are often stored in `spec/fixtures`, `features/fixtures`, or `test/fixtures`.
Does it make sense?

mattwynne:
OK, understood. Thanks for the explanation.

jarl-dk:
@mattwynne : As you can imagine the example
`check_binary_file_content('produced_file.jpeg','fixtures/expected_file_content.jpeg')` will not really work because fixtures are stored in `spec/fixtures`, `features/fixtures`, or `test/fixtures` and the above file paths (especially) the second parameter are relative to current working directory which is a new directory in every run (under `tmp/aruba`. So the purpose of this PR is to introduce a special character (just like `~` for HOME) that expands to the directory with fixtures. This specific PR proposes to use `%` for that turning the example into `check_binary_file_content('produced_file.jpeg','%/expected_file_content.jpeg')`, but other characters have been proposed (such as `-`, `|`, `^`, `>`) starting at https://github.com/cucumber/aruba/pull/219#issuecomment-66425190
Personally I prefer a single character and `|`, `<`, and `>` may be confused with shell piping.

jarl-dk:
Actually I have often used fixtures (images) to test web-applications using cucumber. So in some sense I see the fixture expansion feature (the purpose of this PR) to be more generally useful than just for CLI testing. So someday I see this feature being implemented in a more general ruby gem (cucumber or even better: mini-test). But maybe we should just start here in Aruba to demonstrate proof-of-concept.

lib/aruba/api.rb
```diff
@@ -729,6 +735,61 @@ def io_wait
       @aruba_io_wait_seconds || DEFAULT_IO_WAIT_SECONDS
     end
 
+    DEFAULT_ROOT_DIRECTORY = Dir.getwd
+
+    # The root directory of aruba
+    def root_directory
```
jarl-dk: Making aruba root directory configurable is not related to this PR (as I see it). I suggest you take that out in a seperate PR.

lib/aruba/api.rb
```diff
@@ -729,6 +735,61 @@ def io_wait
...
+      @aruba__root_directory ||= DEFAULT_ROOT_DIRECTORY
+    end
+
+    DEFAULT_FIXTURES_PATH_PREFIX = '%/'
+
+    def fixtures_path_prefix
```
jarl-dk: Making the fixture prefix character configurable via `@aruba_fixture_path_prefix`. To be honest I think it should not be configurable; just like `~` referring to HOME directory is not configurable.

lib/aruba/api.rb
```diff
@@ -729,6 +735,61 @@ def io_wait
...
+        features/fixtures
+        test/fixtures
+      ).map { |dir| File.join(root_directory, dir) }.find(if_not_found) { |dir| File.directory? dir }
+    end
+
+    # Return known fixtures
+    #
+    # @return [Array]
+    #   An array of known files and directories in fixtures directory
+    def known_fixtures
```
jarl-dk: I consider this to be feature creep. Stick to the simplest possible solution first. Then open another PR for this feature and argue why this is necessary in Aruba.

lib/aruba/api.rb
```diff
@@ -729,6 +735,61 @@ def io_wait
...
+          relative_path: Pathname.new(path).relative_path_from(Pathname.new(fixtures_directory)).to_s
+        )
+      end
+    end
+
+    # Find a fixture among known ones
+    #
+    # @return [String]
+    #   An array of known files and directories in fixtures directory
+    def find_fixture(pattern)
```
jarl-dk: This also seems like feature creep to me. Am I missing something?

lib/aruba/api.rb
```diff
@@ -729,6 +735,61 @@ def io_wait
...
+    end
+
+    DEFAULT_FIXTURES_DIRECTORY = 'fixtures'
+
+    # The path to the directory which contains fixtures
+    # You might want to overwrite this method to place your data else where.
+    #
+    # @return [String]
+    #   The directory to where your fixtures are stored
+    def fixtures_directory
```
jarl-dk: I think the finding of the fixture directory can be implemented such that it is only executed once only. I.e. Setting a class variable straight in the class body. And making it configurable like `#exit_timeout` and `#io_wait` Does it make sense?

lib/aruba/api.rb
dg-ratiodata: I've no use case for this method. I'm happy to remove this code. I wrote it for an earlier implementation of this PR and left in this PR, in the case you might find this useful. Sry for not mentioning this.

lib/aruba/api.rb
dg-ratiodata: See above... Same here.

lib/aruba/api.rb
dg-ratiodata: I see, but it makes the configuration of the fixtures more cleaner. I added a new PR for this. Would you mind to merge it first, then?

lib/aruba/api.rb
dg-ratiodata: ok. changed.

dg-ratiodata:
I cleaned up the implementation and rebased it on master... BTW I left the `root_directory`-implementation within this PR, but opened a new PR #232 for adding it. I will rebase this PR when/if #232 is merged.

jarl-dk:
I was a bit fast on criticising `absolute_path`. I see that it was already part of v0.6.2.

I am glad that you take the time to rename it and make `absolute_path` deprecated.

lib/aruba/api.rb
```diff
@@ -21,20 +23,32 @@ module Api
...
+      else
+        in_current_dir { File.expand_path File.join(*args) }
+      end
+    end
+
+    # @private
+    # @deprecated
     def absolute_path(*args)
-      in_current_dir { File.expand_path File.join(*args) }
+      warn('The use of "chmod" is deprecated. Use "filesystem_permissions" instead')
```
jarl-dk: Looks like copy paste code from `chmod`

dg-ratiodata: ups again. Will be fixed...

dg-ratiodata:
@jarl-dk Fixed.

dg-ratiodata:
rebased with `master`. waiting for #232 to be rebased again.

dg-ratiodata:
@jarl-dk @mattwynne Anything missing here? Would love to see this merged soon. Please be aware that this PR includes code from #232. So it makes sense to merge #232 first.

jarl-dk:
I have rework abit, @dg-ratiodata Please review my rework. Still I think this feature is not exploited in feature tests where I think it belongs most. Basically the new `Aruba#expand_path` should be used everywhere in `lib/aruba/cucumber.rb` so the fixtures file path can be used in feature steps. So some more work is needed there...

PR #229, opened 2015-01-23
Support for setting mtime when using #touch_file

dg-ratiodata:
@jarl-dk 

When creating a fixture using `touch_file` it can be quite handy to set the timestamp to a fixed time. This PR makes this possible by passing through options to `FileUtils`.

This

```ruby
touch_file 'file.txt', mtime: Time.parse('2014-01-01 10:00:00')
```

works like that

```bash
#/bin/bash
touch -d '2014-01-01 10:00:00' file.txt
```
This also speeds up tests, because you do not need to do something nasty like this:

```ruby
touch_file 'file.1.txt'
sleep 1
touch_file 'file2.txt'
```

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -432,13 +432,10 @@ module Aruba
     # @param [true, false] expect_match
     #   Must the content be in the file or not
     def check_binary_file_content(file, reference_file, expect_match = true)
-      prep_for_fs_check do
-        identical = FileUtils.compare_file(file, reference_file)
-        if expect_match
-          expect(identical).to be(true), "The file \"#{file}\" differs from file \"#{reference_file}\""
-        else
-          expect(identical).not_to be(true), "The file \"#{file}\" is identical to file \"#{reference_file}\""
-        end
+      if expect_match
+        expect(file).to have_same_file_content_like reference_file
+      else
+        expect(file).not_to have_same_file_content_like reference_file
       end
     end
 
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -96,13 +96,19 @@ module Aruba
     #
     # @param [String] file_name
     #   The name of the file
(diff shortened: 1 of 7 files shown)


dg-ratiodata:
@jarl-dk 

This branch now contains all changes for  #220 and #218. It was a lot of easier for me to add them here than resolving merge conflicts.

I also refactored some tests by using "patterns" found in this [talk](https://www.youtube.com/watch?v=d2gL6CYaYwQ) along my way through the code, I added some more tests and added some matchers to "clarify" the intent of the code. I left all other tests untouched.

I think we could clean up some more specs and replace them by "shared examples", but I leave them now as they are. Maybe we can change later them in a separate PR.

Maybe we can replace some more "check_*"-methods by implementing `custom matchers` instead, but I think this should be discussed in a separate PR/issue.

I only changed those methods, which make use of `FileUtils`. All changes made do not change the existing API. I know that there are several changes mixed up in this PR, but I did my best to separate them in different commits. 

Cheers...

dg-ratiodata:
Rebased...

maxmeyer:
This PR has been opened for a while. 

@aslakhellesoy @mattwynne @jarl-dk Is there anyone willing to review this?

mattwynne:
It seems to me these changes make a positive overall contribution to the codebase and should be merged. @jarl-dk do you agree?

jarl-dk:
Sorry for being away for some time.

Thanks for notifying me.

I'll take a look at all this today or tomorrow.

maxmeyer:
Any comments? :smile:

jarl-dk:
Thanks for all this work. It would be nice to have som more ruby doc on the arguments now that many methods just takes a `args*` argument with no hint of what is possible.

dg-ratiodata:
ok.

dg-ratiodata:
Thanks for merging.

PR #232, opened 2015-04-07
Feature/root directory

dg-ratiodata:
PR based on comment from @jarl-dk on this PR #224.

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -788,6 +788,13 @@ module Aruba
       @aruba_io_wait_seconds || DEFAULT_IO_WAIT_SECONDS
     end
 
+    DEFAULT_ROOT_DIRECTORY = Dir.getwd
+
+    # The root directory of aruba
+    def root_directory
+      @aruba_root_directory ||= DEFAULT_ROOT_DIRECTORY
+    end
+
     # Run a command with aruba
     #
     # Checks for error during command execution and checks the output to detect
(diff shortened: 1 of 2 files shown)


dg-ratiodata:
Failure no travis seems to be a bug in cucumber, see cucumber/cucumber#827 for this.

dg-ratiodata:
rebased.

dg-ratiodata:
@jarl-dk @mattwynne What about this one? This provides an helper to get the root directory of `aruba`. I used this helper in #224 to find the fixtures directory.

By default this is `Dir.getwd`. But using this directly is not enough because CWD can be changed by using [`cd`](https://github.com/cucumber/aruba/blob/master/lib/aruba/api.rb#L88) later on.

dg-ratiodata:
Merged.

mattwynne:
I know this is consistent with other configuration settings like `@aruba_io_wait_seconds` but it reminds me that I don't like this mechanism of configuration by setting instance variables. 

I'd like us to go over to using `Aruba.configure` for all this stuff.

dg-ratiodata:
Sounds good. Something like `RSpec.configure`?

mattwynne:
Yes, exactly like that. The [mechanism already exists](https://github.com/cucumber/aruba/blob/master/lib/aruba/config.rb#L37) in Aruba, but it's not used consistently.

mattwynne:
Can you create a new ticket for that @maxmeyer?

mattwynne:
Or should I say, @dg-ratiodata? ;)

dg-ratiodata:
:-) Either is ok....

PR #233, opened 2015-04-07
Use 1.x for cucumber to work around bug

dg-ratiodata:
Using `cucumber` "2.x" make tests fail for aruba. It looks like `cucumber` 2.x + `gherkin` 2.x doesn't work for the test suite of aruba. It looks like we should wait for `gherkin` 3 (see @aslakhellesoy [comment](https://github.com/cucumber/cucumber/issues/827#issuecomment-90483021) on this. This PR pins `aruba` to `cucumber` 1.x. The commit should be reverted when `gherkin` 3 is out.

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
   s.email       = 'cukes@googlegroups.com'
   s.homepage    = 'http://github.com/cucumber/aruba'
 
-  s.add_runtime_dependency 'cucumber', '>= 1.1.1'
+  s.add_runtime_dependency 'cucumber', '~> 1.0'
   s.add_runtime_dependency 'childprocess', '>= 0.3.6'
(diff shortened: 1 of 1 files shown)


mattwynne:
I would like to understand why Cucumber 2.0 doesn't work. It seems very odd to me that Cucumber 2.0 has suddenly exposed a new bug in Gherkin.

aslakhellesoy:
It should be possible to fix this without waiting for Gherkin3 (which will take at least several months).

The latest versions of cucumber 1.x and 2.x use the same version of gherkin 2.x. I don't know why cucumber 2.x causes gherkin to blow up like this...

dg-ratiodata:
Would it be ok, that you guys work that out?

dg-ratiodata:
Digging into stack traces goes beyond my skills.

jarl-dk:
I suggest we merge this pinning until problem is solved in cucumber.

PR #234, opened 2015-04-08
Make sure a file/directory does not exist + Cleanup named file/directory steps

dg-ratiodata:
I often use an `init`-command to create some files. In my steps I try to make it obvious that the file should not exist. This PR adds new cucumber steps which makes the intent very clear and does not fail if I file/directory does not exist. This PR also includes a small refactoring - hope this is ok.

New steps
* Make sure file does not exist
* Make sure directory does not exist

Cleanup step
* Make directory step look the same like file step

I'm very well aware of the `I remove file/directory "xz.txt"`-step. But I think the intention is very different.

### DIFF
--- a/features/file_system_commands.feature
+++ b/features/file_system_commands.feature
@@ -17,6 +17,14 @@ Feature: file system commands
     When I run `cat foo/bar/example.txt`
     Then the output should contain exactly "hello world"
 
+  Scenario: a file does not exist
+    Given a file named "example.txt" does not exist
+    Then the file "example.txt" should not exist
+
+  Scenario: a directory does not exist
+    Given a directory named "example.d" does not exist
+    Then the directory "foo" should not exist
+
   Scenario: create a fixed sized file
     Given a 1048576 byte file named "test.txt"
     Then a 1048576 byte file named "test.txt" should exist
@@ -125,7 +133,7 @@ Feature: file system commands
     Given a directory named "foo/bar"
     Then the following step should fail with Spec::Expectations::ExpectationNotMetError:
       """
-      Then a directory named "foo/bar" should not exist
+      Then the directory "foo/bar" should not exist
       """
 
   Scenario: Check file contents with text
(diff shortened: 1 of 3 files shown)


dg-ratiodata:
Mmh... Seems like there are some more inconsistencies in step names:

```
Then /^a file named "([^"]*)" should exist$/ do |file|
  check_file_presence([file], true)
end

Then /^a file named "([^"]*)" should not exist$/ do |file|
  check_file_presence([file], false)
end

Then /^the file "([^"]*)" should not exist$/ do |file_name|
  check_file_presence([file_name], false)
end
```

@jarl-dk What do you think which wording should be used? I don't think we should support both ones.

mattwynne:
> I don't think we should support both ones.

Actually, I'd love it if we did. I keep forgetting the 'named' bit when I use these steps in Cucumber's features, and I've always thought it read awkwardly. What about making it consistently optional?

dg-ratiodata:
> I keep forgetting the 'named' bit when I use these steps in Cucumber's features,

Me too

> What about making it consistently optional?

What do you mean with "consistently optional"?

mattwynne:
Something like this:

```
Then /^a file(?: named)? "([^"]*)" should exist$/ do |file|
  check_file_presence([file], true)
end
```

dg-ratiodata:
ah. ok. I see. will fix that.

dg-ratiodata:
Ok. I changed some tests to test the new behaviour. Now all steps for directories/files are built like that. I also change the "not"-thing as well to reduce the number of needed steps - a pattern I found in other existing steps.

```ruby
Then /^(?:a|the) file(?: named)? "([^"]*)" should (not )? exist$/ do |file, expect_match|
  check_file_presence([file], !expect_match)
end
```

dg-ratiodata:
If you prefer I can split up this PR and make a new one for the newly added steps.

dg-ratiodata:
@mattwynne @jarl-dk 

ok?

mattwynne:
LGTM. It might be worth rebasing your commits to keep the history tidy when we merge this in.

dg-ratiodata:
Rrebased.

dg-ratiodata:
@mattwynne Can we go on with this PR?

dg-ratiodata:
Rebased.

PR #237, opened 2015-04-28
Add matcher to check if file exists

dg-ratiodata:
This adds a helper to check if a file exists in filesystem. This also resolves the full path. ~~This PR needs to be rebased if #224 is merged (absolute_path -> expand_path).~~

I added a separate spec file. This spec file uses the aruba api as well. To make the "aruba api object" `@aruba` re-usable I added a shared context for that.

### DIFF
--- a/spec/support/shared_contexts/aruba.rb
+++ b/spec/support/shared_contexts/aruba.rb
@@ -1,0 +1,32 @@
+RSpec.shared_context 'uses aruba API' do
+  def random_string(options = {})
+    options[:prefix].to_s + SecureRandom.hex + options[:suffix].to_s
+  end
+
+  before(:each) do
+    klass = Class.new {
+      include Aruba::Api
+
+      def set_tag(tag_name, value)
+        self.instance_variable_set "@#{tag_name}", value
+      end
+
+      def announce_or_puts(*args)
+        p caller[0..2]
+      end
+    }
+    @aruba = klass.new
+
+    @file_name = "test.txt"
+    @file_size = 256
+    @file_path = File.join(@aruba.current_directory, @file_name)
+
+    (@aruba.dirs.length - 1).times do |depth| #Ensure all parent dirs exists
+      dir = File.join(*@aruba.dirs[0..depth])
+      Dir.mkdir(dir) unless File.exist?(dir)
+    end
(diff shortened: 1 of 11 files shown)


dg-ratiodata:
@jarl-dk @mattwynne Suggestions?

dg-ratiodata:
Mmmh... There are more commits here now than I expected, when I started improving this PR. Hope this is not too much. Otherwise I will try to split it up.

spec/support/shared_contexts/aruba.rb
```diff
@@ -0,0 +1,32 @@
...
+      include Aruba::Api
+
+      def set_tag(tag_name, value)
+        self.instance_variable_set "@#{tag_name}", value
+      end
+
+      def announce_or_puts(*args)
+        p caller[0..2]
+      end
+    }
```
mattwynne: This feels a bit messy. Is it necessary to define a new instance of this `klass` per example? Could you not just define an `ArubaRunner` class or somesuch once, then use instances of that for `@aruba`?

spec/support/shared_contexts/aruba.rb
```diff
@@ -0,0 +1,32 @@
...
+    @aruba = klass.new
```
mattwynne: I would prefer to use a `let` block to define `aruba` then use that (without the `@`) in the specs that depend on it.

mattwynne:
Thanks @dg-ratiodata, the new helpers and matchers look great!

spec/support/shared_contexts/aruba.rb
maxmeyer: Agreed. I would prefer that too. Is it ok, if I open a new PR for that? I think we need to update the rest of the test suite for that.

spec/support/shared_contexts/aruba.rb
maxmeyer: Mmmh... Maybe this should be done in another PR. I only move the (old) code.

mattwynne: Oh I'm sorry I didn't realise! Just go ahead then, and yes we can do that in another PR.

spec/support/shared_contexts/aruba.rb
mattwynne: Yes please!

maxmeyer:
Alreday merged by #238

PR #238, opened 2015-04-28
Added matcher for file content

dg-ratiodata:
I added a matcher to match string to file content. This might be useful for aruba internals as well, but is itended to be part of the public API.

This PR is based on a refactoring done in #237, that's why there are some more commits added here. I will rebase this PR.

### DIFF
--- a/lib/aruba/matchers/file.rb
+++ b/lib/aruba/matchers/file.rb
@@ -115,3 +115,36 @@ RSpec::Matchers.define :have_file_content do |expected|
     format("expected that file \"%s\" does not contains exactly:\n%s", actual, expected)
   end
 end
+
+# @!method have_file_size(size)
+#   This matchers checks if path has file size
+#
+#   @param [Fixnum] size
+#     The size to check
+#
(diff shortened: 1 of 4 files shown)


dg-ratiodata:
@jarl-dk @mattwynne Suggestions? I split this up in two PRs. This is why this one depends on changes in #237

maxmeyer:
I added another matcher... :smile:

mattwynne:
It looks to me like the `have_file_content` matcher will not show a diff if the contents of two files are different. Our string comparison (e.g. for stdout) does this, and it's very handy.

mattwynne:
I suggest adding that in another iteration.

dg-ratiodata:
I learned yesterday about "diffable" in a matcher. Will add this.

mattwynne:
Cool!

You know, being really picky, it strikes me that the `if` statement checking the type of the expected value (Regexp or String) appears in three places in the matcher, and it would be better to use polymorphism, and just have two different matcher objects, with the factory method picking the right one.

I'll see if I can show you what I mean.

PR #239, opened 2015-04-28
Make matcher have_permisions public and add documentation

dg-ratiodata:
I think the `have_permissions`-matchers is usable for other people as well. I think we should remove the `@private`-flag.

```ruby
describe 'my feature' do
  it { expect('file.txt').to have_permissions(0700) }
end
```

I also added a configuration file for `yardoc` and the directory which is generated by `yardoc` during documentation generation is ignored now.

@mattwynne @jarl-dk Suggestions?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ cucumber-pro.log
 
 ## Ruby versioniser
 .ruby-version
+
+# Temp files of yar
+.yardoc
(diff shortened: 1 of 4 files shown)


dg-ratiodata:
Rebased.

mattwynne:
LGTM

PR #240, opened 2015-04-28
Improve development environment

dg-ratiodata:
1. Remove ruby 1.9.3 from supported rubies

  Ruby 1.9.3 is End of Life. I see no reason to officially support a ruby
  version which does not get any bugfixes from core developers.

2. Add support for JRuby 2 and 2.1 mode

3. Debugging

  `pry` seems to be the debugger-tool in the ruby-world. Together with byebug
  (ruby 2 only) even code walking is possible. The stack_explorer plugin gives
  you a nice access to the stack trace. The doc-plugin makes ruby doc available from within a pry session

4. License Finder

  `aruba` seems be used widely. So extra care should be taken with licensing
  issues. The [license_finder](https://github.com/pivotal/LicenseFinder) helps
  you with that. It tracks license of all uses libraries. I whitelisted the MIT
  and the BSD-licenses, I think there's no problem here. I excluded the gems
  which are part of the `development`-group. They are not needed for the
  core-functionality.

@jarl-dk @mattwynne  Feedback?

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,0 +1,8 @@
+SimpleCov.start do
+  add_filter "/features/"
+  add_filter "/spec/"
+  add_filter "/tmp"
+  add_filter "/vendor"
+  
+  add_group "lib", "lib"
+end
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: false
 language: ruby
+bundler_args: --without development debug profile
+script: script/test
 rvm:
 - 2.1.5
 - 2.1.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: ruby
-bundler_args: --without development debug profile
 script: script/test
+install: script/bootstrap --without development debug
 rvm:
 - 2.1.5
 - 2.1.4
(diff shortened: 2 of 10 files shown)


mattwynne:
This is great. I suggest you update https://github.com/cucumber/aruba/blob/master/CONTRIBUTING.md to explain how to use the bootstrap and test scripts, then merge it in.

mattwynne:
Thinking about it, I'm not sure why we need `script/test` - I think the default should be just to run `bundle exec rake`

dg-ratiodata:
There was a talk from a guy from github. Using `script/*` can be used in projects using different languages whereas `bundle exec rake` is ruby specific.

dg-ratiodata:
It's more a convience thing, that other developers do not need to think about running `rake test` or `rake test:rspec` or ... but hey there's a /script/test let's run this script to run tests.

mattwynne:
OK well I don’t mind it. My fingers will always tend to run `rake` in a Ruby project in any case.

dg-ratiodata:
Rebased, added documentation and make use of the bootstrap script on travis to make sure it really works.

mattwynne:
Go ahead and merge it then Dennis!

dg-ratiodata:
Access Denied.

maxmeyer:
Ok. Done.

mattwynne:
\o/

PR #241, opened 2015-04-30
use expand path everywhere

dg-ratiodata:
Follow up to #224.

@jarl-dk 

You changed the public API of `#absolute_path`. I think we should deprecate that API properly first. To make the switch easier for us I added the old implementation of `absolute_path` and left the new one to `expand_path`. What do you think?

To make it possible to use `expand_path` everywhere I removed the `prep_for_fs`-call, because otherwise we would have doubled the `tmp/aruba`-path in the expanded path.

Do you see any reason to keep the method as part of `aruba`. I mark it as deprecated.

I added single commits for the changes to make review easier. I will squash them after the review.

I also added a spec for the `absolute_path` as long as it is part of `aruba`. The deprecation warning is silenced.

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -474,12 +474,12 @@ module Aruba
     # @yield
     #   Pass the content of the given file to this block
     def with_file_content(file, &block)
-      prep_for_fs_check do
-        file = File.expand_path(file)
+      stop_processes!
 
-        content = IO.read(file)
-        yield(content)
-      end
+      file = expand_path(file)
+      content = IO.read(file)
+
+      yield(content)
     end
 
     # Check the content of file
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -410,9 +410,9 @@ module Aruba
     # @param [true,false] expect_presence
     #   Should the given paths be present (true) or absent (false)
     def check_file_presence(paths, expect_presence = true)
-      stop_processes!
-
       warn('The use of "check_file_presence" is deprecated. Use "expect().to be_existing_file or expect(all_paths).to match_path_pattern() instead" ')
+
+      stop_processes!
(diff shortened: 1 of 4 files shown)


dg-ratiodata:
@mattwynne 

Rebased

mattwynne:
Seems reasonable to me, but why is the `Reporting` module added in here?

dg-ratiodata:
It's not aruba's `Reporting`-module. The code is needed to silence some deprecation warnings, during testing deprecated API-methods as long as they are part of aruba.

See spec/aruba/api_spec.rb:352 of PR.

dg-ratiodata:
https://github.com/cucumber/aruba/pull/241/files#diff-b014353d9f47f067ea8e078d34a1826bR352

maxmeyer:
@mattwynne 

Ok? Any other suggestion? Will then squash commits to three or four and merge those then.

jarl-dk:
Thanks @maxmeyer . Obviously my work was a bit rushed. You've cleaned up quite exactly what I wanted to do (and more) but could not find time to do... Thanks... Next thing I would love to see is to allow `%` in all path arguments, basically that means using `Aruba#expand_path` in stead of File#expand_path. That is not done completely by https://github.com/dg-ratiodata/aruba/commit/0ebd0832e3bf98898d6ba6c7c9845ae01dc8ff28, right?

Anyway: :+1: go merge.

maxmeyer:
Thanks.

maxmeyer:
@jarl-dk 

At some places that doesn't work. You got an recursive invocation.

maxmeyer:
Merged.

jarl-dk:
@maxmeyer, Yes, I realised when developing. But I didn't finish it. The recursive issue calls for a clean-up.

PR #244, opened 2015-05-02
Add rspec integrator

maxmeyer:
Make it easier for others to integrate `aruba` with rspec.

BTW and as a reminder for me in the future:
This https://github.com/jnicklas/capybara/blob/master/lib/capybara/rspec.rb is a good example.

We may need to discuss the `type`. I'm open for suggestions. What about `type: :aruba` or `type: :cli`?


dg-ratiodata:
I also added some code I would like to activate when the new announcer I'm developing is accepted and merged.

Issue #231, opened 2015-04-05
Refactor API


maxmeyer:
Hi @jarl-dk  @mattwynne ,

I would like to discuss the following things with you:

1. Move "Helper"-methods to `aruba/helpers.rb` - e.g. `absolute_path`
2. Move internal helpers to `aruba/utils.rb` and make them `module`-functions of `Aruba::Utils`

   '1.' and '2.' will make clear, what is part of the `aruba`-API

3. Migrate `check_*`-methods to `rspec`-matchers
4. Migrate `assert*`-methods to `rspec`-matchers
   
   '3.' and '4.' will make tests more readable + providing better error methods
   to users. And will make use of all the good things which come with
   `expect(x)`. And as we are already using `rspec` for the api-tests,
   testing of the new `matchers` will be dead simple.

   Alternative: Create more custom matchers to improve code.

   *Example*:

   **old**

   ```ruby
   def check_file_size(paths_and_sizes)
     prep_for_fs_check do
       paths_and_sizes.each do |path, size|
         path = File.expand_path(path)
         expect(File.size(path)).to eq size
       end
     end
   end
   ```

   **new**

   ```ruby
   def check_file_size(paths_and_sizes)
     prep_for_fs_check do
       paths_and_sizes.each do |path, size|
         path = File.expand_path(path)
         expect(path).to have_file_size size
       end
     end
   end
   ```

   Writing `rspec` tests using the `assert_*` looks very weird. I would prefer
   to have custom `rspec`-matchers for this.

   **old**

   ```ruby
   def assert_partial_output(expected, actual)
     actual.force_encoding(expected.encoding) if RUBY_VERSION >= "1.9"
     expect(unescape(actual)).to include(unescape(expected))
   end
   ```

   **new**

   ```ruby
   def assert_partial_output(expected, actual)
    $stderr.puts 'Deprecated'
     expect(actual).to include_content expected
   end

   # matcher
   RSpec::Matchers.define :include_content do |expected|
     match do |actual|
       actual.force_encoding(expected.encoding) if RUBY_VERSION >= "1.9"
       
       # Support real files as well
       actual = File.read(absolute_path(actual.to_s)).chomp if actual.is_a? Pathname
       Aruba::Utils.unescape(actual).include? Aruba::Utils.unescape(expected)
     end

     failure_message do |actual|
       format("expected that file \"%s\" has content \"%s\".", actual, expected)
     end

     failure_message_when_negated do |actual|
[...]

PR #245, opened 2015-05-04
Moving code to separate classes

dg-ratiodata:
This is the first PR for #231. I kept the commits as isolated as possible, to make it easier to review the changes.

This PR is not yet finished. I think it needs to be discussed first. All "old" tests are passing with this code. It also duplicates the `capture`-method I used in some other PR. I only added it to make the tests pass. The documentation is not complete. Tests needs to be written.

Besides that I'm quite happy with the code:
* It support's the old API and adds a lot of deprecation warnings
* Now all announcer code is bundled and has a nice API which be extended easily by us and by users - see a very first version of an example in `lib/aruba/announcer.rb`
* The `announce_or_puts`-method is gone and replaced by `mode=`
* I introduced a `default_announcer` which writes to `/dev/null` and an announcer which writes to what ever announcer was chosen by `mode=` - e.g. `Kernel.puts`, `puts`
* At first I thought it might be a good idea to store the activated announcers in a `Hash`, but decided against it because we would end up with different activated `Announcers` if the mode is changed between two activations
* All process monitoring code is bundled in `ProcessMonitor` - BTW I'm open to rename if one of you guys has a better idea.

@jarl-dk @mattwynne 

Suggestions?


jarl-dk:
I don't see the relation to the topics mentioned in #231. But never mind.

Regarding this PR: It seems like you have seen some light that haven't reached me yet and it will take some time for me to get there... I will just say that at first glance it all seems ok, for the rest of it I'll just trust you..

dg-ratiodata:
Merged into #257

PR #248, opened 2015-05-07
Use polymorphism for file content matcher

mattwynne:


lib/aruba/matchers/file.rb
```diff
@@ -63,34 +63,96 @@
...
-#     end
-#
-#   @example Use matcher with regexp
-#
-#     RSpec.describe do
-#       it { expect(file1).to have_file_content(/a/) }
-#     end
+module RSpec
+  module Matchers
+    # @!method have_file_content(content)
```
dg-ratiodata: I think you can remove `@!method ...`.

dg-ratiodata:
I like the idea a lot. :smile:

lib/aruba/matchers/file.rb
```diff
@@ -63,34 +63,96 @@
...
+    #       it { expect(file1).to have_file_content('a') }
+    #     end
+    #
+    #   @example Use matcher with regexp
+    #
+    #     RSpec.describe do
+    #       it { expect(file1).to have_file_content(/a/) }
+    #     end
+    def have_file_content(expected)
+      if expected.is_a? Regexp
```
dg-ratiodata: I think we could refactor this to something like this

```ruby
@matchers = []
@matchers <<  Aruba::Matchers::HaveExactFileContent
@matchers << Aruba::Matchers::HaveMatchingFileContent

class Aruba::Matchers::HaveMatchingFileContent
  def self.handles?(input)
    input.is_a? Regexp
  end
end

# Handles everything else => always true
class Aruba::Matchers::HaveExactFileContent
  def self.handles?(input)
    true
  end
end

@matchers.find { |m| m.handles? expected }
```

dg-ratiodata:
Should we use this kind of pattern for all other matchers as well? Or should we use the DSL for simpler ones?

dg-ratiodata:
@jarl-dk @mattwynne 
Opinions to that?

myronmarston:
RSpec 3.x provides much improved support for composable matchers that means you don't have to worry about string vs regexp.  I'll try opening a PR showing what I mean in a bit.

lib/aruba/matchers/file.rb
mattwynne: very nice!

mattwynne: remember you can just push commits to this branch now you've got the karma

mattwynne:
No need to mention me @dg-ratiodata, I get notified of each comment in this repo anyway.

Let's see what @myronmarston comes up with, thanks Myron!

myronmarston:
See #249.

jarl-dk:
This looks cool, though superseded by #249 IMHO. The PR needs som rubocop work, but then again if we are heading towards #249 maybe just leave this.

mattwynne:
Indeed, #249 is the one.

PR #249, opened 2015-05-07
Matcher improvements

myronmarston:
This is an alternative to #248.  It uses RSpec 3's support for composable matchers via `values_match?` and `description_of` for the description/failure messages.  The failure messages are derived from the description so we can just set that.  By using `values_match?` there's no need to treat the regexp, string or any other case as special.

Docs:

http://rspec.info/documentation/3.2/rspec-expectations/RSpec/Matchers/Composable.html#values_match%3F-instance_method

http://rspec.info/documentation/3.2/rspec-expectations/RSpec/Matchers/Composable.html#description_of-instance_method

Bear in mind that this will only work on RSpec 3+.  What versions of RSpec is aruba designed to support?  It may need to be rejiggered to support older versions of RSpec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,8 @@ pkg
 
 ## PROJECT::SPECIFIC
 tmp
+bin
+bundle
 .bundle
 doc
 Gemfile.lock
--- a/lib/aruba/matchers/file.rb
+++ b/lib/aruba/matchers/file.rb
@@ -64,21 +64,19 @@ RSpec::Matchers.define :be_existing_file do |_|
 end
 
 # @!method have_file_content(content)
-#   This matchers checks if <file> has content. It checks exactly if `'string'`
-#   is given and does a substring check if `/regexp/` is given.
+#   This matchers checks if <file> has content. `content` can be a
+#   string, regexp or an RSpec matcher.
 #
-#   @param [String, Regexp] content
-#     The content of the file
+#   @param [String, Regexp, Matcher] content
+#     Specifies the content of the file
 #
 #   @return [TrueClass, FalseClass] The result
 #
 #     false:
 #     * if file does not exist
-#     * if file content is not equal string
-#     * if file content does not include regexp
+#     * if file content does not match the content specification
 #     true:
-#     * if file content includes regexp
-#     * if file content is equal string
+#     * if file content matches the content specification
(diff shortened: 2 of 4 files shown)


dg-ratiodata:
We're using `expect()` everywhere. Is this supported by `RSpec < 3`?

mattwynne:
Wow, thanks Myron that's very clean. I don't think we need to care about < RSpec 3 support.

myronmarston:
BTW, if you're curious how the `values_match?` logic works, 

> We're using expect() everywhere. Is this supported by RSpec < 3?

`expect` was introduced in RSpec 2.11, nearly two years before RSpec 3 was released:

http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/

jarl-dk:
Looks very cool.

maxmeyer:
@jarl-dk @myronmarston 
Thanks a lot. Will have a look at it.

dg-ratiodata:
Hi @myronmarston I would like to use the suggested `values_match?`-helper in a matcher which should be diffable - on another project.

```ruby
RSpec::Matchers.define :have_option_value do |expected|
  match do |actual|
    values_match? subject.public_send(actual.to_sym), expected
  end

  diffable

  failure_message do |actual|
    format(%(expected that option "%s" has value "%s"), actual, expected)
  end

  failure_message_when_negated do |actual|
    format(%(expected that option "%s" does not have value "%s"), actual, expected)
  end
end
```

The diff should be done by using the value of `subject.public_send(actual.to_sym)` leaving the `actual`-value untouched because I need it for the failure messages. 

Is there a way to express this using the dsl? While reading the rspec sources I didn't find a proper way to do it. Thanks a lot in advance.

myronmarston:
@dg-ratiodata -- yes, there is a way to do that (and there's also a suggestion I have for your use of `values_match?`) but an aruba issue isn't the right forum for that discussion.  Can you open an rspec-expectations issue, /cc me and I'll be happy to respond there?

dg-ratiodata:
ok

dg-ratiodata:
done.

maxmeyer:
@myronmarston 

Thanks a lot for your suggestions. They were very helpful. I added them to the existing code.

PR #250, opened 2015-05-07
Path Content

dg-ratiodata:
This reads content from
* directory
* file

And returns an array of content. Next step would be to use it "everywhere" in the code. `rubcop` complains if I use `ArgumentError` although that should be ok. For now I disabled the check.

### DIFF
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -122,6 +122,22 @@ module Aruba
       list('.').map { |p| expand_path(p) }
     end
 
+    # Return all existing files in current directory
+    #
+    # @return [Array]
+    #   List of files
+    def all_files
+      list('.').select { |p| file? p }.map { |p| expand_path(p) }
+    end
+
+    # Return all existing directories in current directory
+    #
+    # @return [Array]
+    #   List of files
+    def all_directories
+      list('.').select { |p| directory? p }.map { |p| expand_path(p) }
+    end
+
     # @deprecated
     # @private
     def in_current_dir(*args, &block)
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -98,6 +98,22 @@ module Aruba
       File.directory? expand_path(file)
     end
 
+    # Check if path is absolute
+    #
+    # @return [TrueClass, FalseClass]
+    #   Result of check
+    def absolute?(path)
+      Pathname.new(path).absolute?
+    end
+
+    # Check if path is relative
+    #
+    # @return [TrueClass, FalseClass]
+    #   Result of check
+    def relative?(path)
+      Pathname.new(path).relative?
+    end
+
     # Return all existing paths (directories, files) in current dir
     #
     # @return [Array]
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -103,7 +103,7 @@ module Aruba
     # @return [Array]
     #   List of files and directories
     def all_paths
-      Dir.glob(expand_path('**/*'))
+      read('.').map { |p| expand_path(p) }
     end
 
     # @deprecated
--- a/lib/aruba/api.rb
+++ b/lib/aruba/api.rb
@@ -171,7 +171,7 @@ module Aruba
     # Return content of directory
     #
     # @return [Array]
-    #   The content of file or directory
+    #   The content of directory
     def list(name)
       fail ArgumentError, %(Path "#{name}" does not exist.) unless exist? name
       fail ArgumentError, %(Only directories are supported. Path "#{name}" is not a directory.) unless directory? name
@@ -185,7 +185,7 @@ module Aruba
     # Return content of file
     #
     # @return [Array]
-    #   The content of file or directory
+    #   The content of file
     def read(name)
       fail ArgumentError, %(Path "#{name}" does not exist.) unless exist? name
       fail ArgumentError, %(Only files are supported. Path "#{name}" is not a file.) unless file? name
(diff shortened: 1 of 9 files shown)


maxmeyer:
Ok. I liked the name `read_content` better. At first I used `content` but that doesn't work because rspec use this as well.

jarl-dk:
I think reading from a file is a fundamentally different thing that getting a list of files and directories for a given path. I don't like the two responsibilities collapsed in one method. I prefer single responsibility principle and separation of concern.

maxmeyer:
So you would prefer to have something like `read_directory` and `read_file`?

mattwynne:
I agree with @jarl-dk. How about `ls(path)` and `read(file)`?

maxmeyer:
Great. Good idea. Will fix this.

mattwynne:
ooh, or even maybe `cat(file)`?

maxmeyer:
aaah... we should decide how we name commands... ;-) `ls` or `list`. For me `read` is ok.

maxmeyer:
... and one thing I forgot... `read` does not concatenate files which `cat` does...

maxmeyer:
So I prefer to have 3 separate commands:

* read - read file content
* list - list directory content
* concatenate - concat file content (using read in a loop)

Since all other commands are written in full, I would prefer list written in full as well.

mattwynne:
Sure, do it. I just though it would be fun to emulate bash commands.

jarl-dk:
How about instead of having `list` for directories, then having a `#dir` to return a `Dir` object. Then the user could do `dir("%/subdir").each {|entry| handle_it(entry)}`.

maxmeyer:
> Sure, do it. I just though it would be fun to emulate bash commands.

@mattwynne My first thought was "cool, like this", but then "what if we decided to implement it differently then the shell command is implemented"

> How about instead of having list for directories, then having a #dir to return a Dir object. Then the user could do dir("%/subdir").each {|entry| handle_it(entry)}.

@jarl-dk I like the idea, but would say it's another command and could be used internally as well.

I know, that a custom matcher would be better, but this reads good as well, doesn't it?

```ruby
it { expect(list('dir.d')). to include 'file.txt' }
```

maxmeyer:
BTW:
Is this normal, that I can change YOUR comments?

mattwynne:
> Is this normal, that I can change YOUR comments?
> 
I think it’s a perk of being a project contributor. I sometimes use it to tidy up when people don’t know how to use markdown properly etc.

dg-ratiodata:
@mattwynne This one, too. Please.

lib/aruba/api.rb
```diff
@@ -126,6 +168,31 @@ def clean_current_dir(*args, &block)
       clean_current_directory(*args, &block)
     end
 
+    # Return content of directory
+    #
+    # @return [Array]
+    #   The content of file or directory
```
mattwynne: Isn't this comment wrong now?

lib/aruba/api.rb
```diff
@@ -126,6 +168,31 @@ def clean_current_dir(*args, &block)
...
+      existing_files            = Dir.glob(expand_path(File.join(name, '**', '*')))
+      current_working_directory = Pathname.new(expand_path('.'))
+
+      existing_files.map { |d| Pathname.new(d).relative_path_from(current_working_directory).to_s }
+    end
+
+    # Return content of file
+    #
+    # @return [Array]
+    #   The content of file or directory
```
mattwynne: Isn't this comment wrong now?

lib/aruba/api.rb
```diff
@@ -98,12 +98,54 @@ def directory?(file)
...
+    #   Result of check
+    def relative?(path)
+      Pathname.new(path).relative?
+    end
+
     # Return all existing paths (directories, files) in current dir
     #
     # @return [Array]
     #   List of files and directories
     def all_paths
```
mattwynne: Why do we need this method? If we have `list` and `expand_path` I think people can compose this themselves. Seems like API noise to me.

mattwynne:
Apart from minor comments this looks great @dg-ratiodata. I really like how you're making the effort to document and spec everything. You're a great asset to the team!

mattwynne:
Please merge this in when you're ready.

lib/aruba/api.rb
maxmeyer: This was a "fun" thing for me :smile?

maxmeyer:
> You're a great asset to the team!

Thanks a lot!

> I really like how you're making the effort to document and spec everything.

Thanks again. I started to refactor the whole feature suite. Will push this soon.

lib/aruba/api.rb
maxmeyer: Besides that, I would like add aruba-helpers for most/all methods of http://ruby-doc.org/core-2.2.0/FileTest.html.

maxmeyer: Oh, I didn't noticed that you mean `all_paths`. This one was added by me to reduce code duplication in `check_file_presence` and in `Then /^(?:a|the) file matching %r<(.*?)> should (not )?exist$/ do |pattern, expect_match|`

lib/aruba/api.rb
maxmeyer: Sure. Fixed.

lib/aruba/api.rb
maxmeyer: Fixed.

maxmeyer:
I leave `all_paths` for now. Otherwise there would be code duplication which I do not like.

PR #257, opened 2015-05-18
Cleanup "Process Management" #1

dg-ratiodata:
This one cleans up `Process Management`. Will create a new PR for the next iteration of cleanup.


maxmeyer:
I think this should be reviewed now. I create a follow-up PR to improve command execution a bit more.

features/debug.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Debug command execution
+
+  As a developer
+  I want to use some debugger in my code and therefor need system() to execute my program
+  In order to find a bug
+
+  @debug
```
maxmeyer: I improved this examples in #260.

maxmeyer:
This is the next candiate to be merged.

README.md
```diff
@@ -185,8 +185,8 @@ Then wire it all up in your `features/support/env.rb` file:
 require 'aruba'
 require 'aruba/in_process'
 
-Aruba::InProcess.main_class = MyMain
-Aruba.process = Aruba::InProcess
+Aruba.process = Aruba::Processes::InProcess
```
dg-ratiodata: There are quite a few classing around starting processes. That's why I namespaced them.

cucumber.yml
```diff
@@ -1,10 +1,9 @@
 <%
 java_version = (RUBY_DESCRIPTION.match(/.*?on.*?(1\.[\d]\..*? )/))[1] if defined?(JRUBY_VERSION)
-std_opts = "--color --tags ~@wip"
-std_opts << " --tags ~@wip-jruby-java-1.6" if defined?(JRUBY_VERSION) && (java_version < '1.7.0')
 
-wip_opts = "--color --tags @wip:3"
-wip_opts = "--color --tags @wip:3,@wip-jruby-java-1.6:3" if defined?(JRUBY_VERSION) && (java_version < '1.7.0')
+std_opts          = "--color --exclude features/fixtures"
```
dg-ratiodata: Otherwise `cucumber` tries to load fixtures with `.rb` at the end.

lib/aruba/announcer.rb
```diff
@@ -0,0 +1,161 @@
+module Aruba
```
dg-ratiodata: Moved announcer to become a first class citizen-Class`.

lib/aruba/announcer.rb
```diff
@@ -0,0 +1,161 @@
...
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def after_init
+      format_string :directory, '$ cd %s'
+      format_string :command, '$ %s'
+      format_string :environment, '$ export %s=%s"'
+
+      # rubocop:disable Metrics/LineLength
+      if @options[:stdout]
```
dg-ratiodata: Those `if`-statements are removed in the upcoming version. I added them to be backwards compatible.

lib/aruba/api.rb
```diff
@@ -677,8 +681,8 @@ def check_directory_presence(paths, expect_presence)
     def prep_for_fs_check(&block)
       warn('The use of "prep_for_fs_check" is deprecated. It will be removed soon.')
 
-      stop_processes!
-      in_current_directory { block.call }
+      process_monitor.stop_processes!
```
dg-ratiodata: For now it's named `process_monitor`. This will be renamed to `command_monitor` or something similar in #259.

lib/aruba/api.rb
```diff
@@ -815,7 +814,7 @@ def assert_no_partial_output(unexpected, actual)
     # @return [TrueClass, FalseClass]
     #   If output of interactive command includes arg1 return true, otherwise false
     def assert_partial_output_interactive(expected)
-      unescape(_read_interactive).include?(unescape(expected)) ? true : false
+      unescape(last_command.stdout).include?(unescape(expected)) ? true : false
```
dg-ratiodata: Running a interactive command is not special anymore. Instead those methods use the last command executed. I think this makes the code more readable and easier to maintain.

lib/aruba/api.rb
```diff
@@ -863,13 +862,16 @@ def assert_exit_status_and_output(expect_to_pass, expected_output, expect_exact_
     #   If arg1 is true, return true if command was successful
     #   If arg1 is false, return true if command failed
     def assert_success(success)
-      success ? assert_exit_status(0) : assert_not_exit_status(0)
+      if success
+        expect(last_command).to be_successfully_executed
```
dg-ratiodata: I also added some more matchers to check the exit status.

lib/aruba/api.rb
```diff
@@ -1036,12 +1049,12 @@ def run_interactive(cmd)
...
     def type(input)
       return close_input if "" == input
-      _write_interactive(_ensure_newline(input))
+      last_command.write(_ensure_newline(input))
     end
 
     # Close stdin
     def close_input
-      @interactive.stdin.close
+      last_command.close_io(:stdin)
```
dg-ratiodata: Nothing special about interactive commands anymore. See one of the comment above.

lib/aruba/api.rb
```diff
@@ -1053,18 +1066,21 @@ def eot
...
-      @interactive.read_stdout
+      warn('The use of "_read_interactive" is deprecated. It will be removed soon.')
+
+      last_command.stdout
     end
 
     # @private
     def _ensure_newline(str)
-      str.chomp << "\n"
+      Utils.ensure_newline(str)
```
dg-ratiodata: I introduced a `Utils`-namespace for methods which are used all over `Aruba`.

lib/aruba/api.rb
```diff
@@ -1076,20 +1092,6 @@ def announce_or_puts(msg)
       end
     end
 
-    # @private
```
dg-ratiodata: This is also move to `Utils`.

lib/aruba/api.rb
```diff
@@ -1129,7 +1131,7 @@ def unset_bundler_env_vars
     # @param [String] value
     #   The value of the environment variable. Needs to be a string.
     def set_env(key, value)
-      announcer.env(key, value)
+      announcer.announce(:environment, key, value)
```
dg-ratiodata: I changed the interface of the announcer to make it easier to extend. Now you just to need `announcer.active :channel1` to use `announcer.announce(:channel1, 'text')`. Maybe we could even make this API public. WDYT?

lib/aruba/api.rb
```diff
@@ -1165,69 +1167,36 @@ def with_env(env = {}, &block)
       restore_env
     end
 
+    # Access to announcer
+    def announcer
+      @announcer ||= Announcer.new(
```
dg-ratiodata: Backward-Compatibility

lib/aruba/api.rb
```diff
@@ -1165,69 +1167,36 @@ def with_env(env = {}, &block)
...
+      @announcer
+    end
+
+    module_function :announcer
+
     # TODO: move some more methods under here!
 
     private
 
     def last_exit_status
```
dg-ratiodata: I think will deprecate this method. Users can now use `last_command.exit_status` to get the exit status of the last command executed.

lib/aruba/cucumber.rb
```diff
@@ -102,6 +102,15 @@
   end
 end
 
+Given /^I append the value to the environment variable:/ do |table|
```
dg-ratiodata: This was a "fun" thing.

lib/aruba/cucumber.rb
```diff
@@ -124,11 +133,11 @@
 
 When /^I run "([^"]*)" interactively$/ do |cmd|
   warn(%{\e[35m    The /^I run "([^"]*)" interactively$/ step definition is deprecated. Please use the `backticks` version\e[0m})
-  run_interactive(unescape(cmd))
```
dg-ratiodata: I think re-using steps is better here.

lib/aruba/cucumber.rb
```diff
@@ -124,11 +133,11 @@
...
+  step %(I run `#{cmd}` interactively)
 end
 
 When /^I run `([^`]*)` interactively$/ do |cmd|
-  run_interactive(unescape(cmd))
+  run(unescape(cmd))
```
dg-ratiodata: `#run` is suitable for this task. No need to have another method for that.

lib/aruba/cucumber/hooks.rb
```diff
@@ -16,35 +16,40 @@
 end
 
 Before('@puts') do
```
dg-ratiodata: I was not sure what the `@puts`-hook is useful for, but now there can be different modes: Null -> Silent, KernelPuts and Puts.

lib/aruba/matchers/command.rb
```diff
@@ -0,0 +1,79 @@
...
+#   @example Use matcher
+#
+#     RSpec.describe do
+#       it { expect(last_command).to have_exit_status(0) }
+#     end
+RSpec::Matchers.define :have_exit_status do |expected|
+  match do |actual|
+    @old_actual = actual
+
+    @announcer ||= Aruba::Announcer.new(
```
dg-ratiodata: I tried to use `announcer` from `Aruba::Api`. This works fine in `RSpec`, but fails in `cucumber`. I'm not sure why, but I think it has something to how the binding is passed to the matcher. See rspec/rspec-expectations#803 for this. Now I use @announcer ... for this.

lib/aruba/process_monitor.rb
```diff
@@ -0,0 +1,113 @@
+module Aruba
```
dg-ratiodata: In this iteration I migrated all methods around starting, stopping, querying information about processes here. In the next iteration we can make more methods private I think.

lib/aruba/processes/basic_process.rb
```diff
@@ -0,0 +1,25 @@
+module Aruba
+  module Processes
+    class BasicProcess
```
dg-ratiodata: Move common code to all processes here.

lib/aruba/processes/debug_process.rb
```diff
@@ -0,0 +1,59 @@
...
+    #
+    # @params [Integer] exit_timeout
+    #   The timeout until we expect the command to be finished
+    #
+    # @params [Integer] io_wait
+    #   The timeout until we expect the io to be finished
+    #
+    # @params [String] working_directory
+    #   The directory where the command will be executed
+    class DebugProcess < BasicProcess
```
dg-ratiodata: This is an interesting one. Flagging a Test with `@debug` activates this launcher. This will make things like `binding.pry` in your cli application possible for a single test to debug a bug occuring during test under cucumber.

```cucumber
@debug
Feature: Blub
  Scenario: Hello World
    Given I run `cucumber`
```

lib/aruba/processes/spawn_process.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+
+      def stderr
+        wait_for_io do
+          @process.io.stderr.flush
+          read(@err)
+        end || @error_cache
+      end
+
+      def read_stdout
```
dg-ratiodata: I don't see a reason why there are two different methods accessing stdout. In the next iteration `def stdout` gets a `wait = true`-parameter to enable/disable waiting for IO to be finished.

script/console
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env ruby
```
dg-ratiodata: This one includes Aruba::Api to make a quick test easier. It opens a pry console on startup.

spec/support/helpers/reporting.rb
```diff
@@ -1,4 +1,4 @@
-module SpecHelper
+module SpecHelpers
   module Reporting
```
dg-ratiodata: This is a different `Reporting`-module. It only includes some code from active support to capture a stream like `$stderr` or `$stdout` from code run in its block.

dg-ratiodata:
@mattwynne What do you think about that PR? It's just the beginning, but I hope this is the right direction. It lays the groundwork to introduce the new config in #258.

lib/aruba/utils.rb
```diff
@@ -0,0 +1,21 @@
+module Aruba
+  module Utils
```
mattwynne: I suggest `Platform` as the name. `Utils` is too general and could become a dumping-ground.

PR #258, opened 2015-05-29
Add configuration

maxmeyer:
This will add a main configuration to aruba. See #246 This one now includes #257 and #219 because I needed some changes I made there.

To discuss:
* [ ] Use of [`contracts`-gem](https://github.com/egonSchiele/contracts.ruby) for configuration
* First Refactored feature test -> More Documentation then test - see features/commands/exit_statuses.feature.
* Relish Support
* Implementation of config
  * There's a global configuration which is dupped to runtime configuration
  * The configuration can be changed in rspec via metadata - does it make sense to port it to cucumber?

PR #259, opened 2015-06-02
Migrate to an observer pattern to inform code about changes going on

dg-ratiodata:
This PR will bring event queues where code can register itself to get informed about changes.

Events
* change of current working directory
* change of environment variables
* new command started
* command stopped
* ...

This is based on PR #257 and #258. It's not finished yet.


dg-ratiodata:
#309 is an improved version of this PR. #309 was merged.

PR #261, opened 2015-06-09
refactor out use_clean_gemset into RVM class

e2:
Also fixes issues:

- now handles newer versions of RVM giving different `rvm gemset create` output

- GEM_PATH was incorrectly replaced (instead preserving @global gemset in path)

### DIFF
--- a/lib/aruba/api/deprecated.rb
+++ b/lib/aruba/api/deprecated.rb
@@ -116,7 +116,7 @@ module Aruba
       #   The command to by run
       #
       # @see #cmd
-      # @deprectated
+      # @deprecated
       def run_interactive(cmd)
         Aruba.platform.deprecated('The use of "#run_interactive" is deprecated. You can simply use "run" instead')
 
--- a/lib/aruba/api/rvm.rb
+++ b/lib/aruba/api/rvm.rb
@@ -15,22 +15,25 @@ module Aruba
       #
       # @param [String] gemset
       #   The name of the gemset to be used
-      def use_clean_gemset(gemset)
-        run_simple(%{rvm gemset create "#{gemset}"}, true)
-        if all_stdout =~ /'#{gemset}' gemset created \((.*)\)\./
-          gem_home = Regexp.last_match[1]
-          set_environment_variable('GEM_HOME', gem_home)
-          set_environment_variable('GEM_PATH', gem_home)
-          set_environment_variable('BUNDLE_PATH', gem_home)
+      def use_clean_gemset(gemset_name)
+        new_gem_home = current_gemset_home(gemset_name)
+        new_bin_path = File.join(new_gem_home, 'bin')
 
-          paths = (ENV['PATH'] || "").split(File::PATH_SEPARATOR)
-          paths.unshift(File.join(gem_home, 'bin'))
-          set_environment_variable('PATH', paths.uniq.join(File::PATH_SEPARATOR))
+        original_gem_home = current_gem_home
+        original_bin_path = File.join(original_gem_home, 'bin')
 
-          run_simple("gem install bundler", true)
-        else
-          raise "I didn't understand rvm's output: #{all_stdout}"
-        end
+        set_environment_variable('GEM_HOME', new_gem_home)
+        set_environment_variable('BUNDLE_PATH', new_gem_home)
+        set_environment_variable('GEM_PATH', replace_paths(ENV['GEM_PATH'], original_gem_home, new_gem_home))
+        set_environment_variable('PATH', replace_paths(ENV['PATH'], original_bin_path, new_bin_path))
+
+        delete_environment_variable('BUNDLE_BIN_PATH')
+        delete_environment_variable('BUNDLE_GEMFILE')
+        delete_environment_variable('RUBYOPT')
+
+        # if not enough, consider 'gem server' and use '--source' in your '~/.gemrc'
+        default_bundler_install_timeout = 30
+        rvm_run("gem install bundler", :timeout => default_bundler_install_timeout)
       end
 
       # Unset variables used by bundler
@@ -38,6 +41,50 @@ module Aruba
         %w[RUBYOPT BUNDLE_PATH BUNDLE_BIN_PATH BUNDLE_GEMFILE].each do |key|
           set_environment_variable(key, nil)
         end
+      end
+
+      private
+
+      def current_gemset_home(gemset_name)
+        output = rvm_capture(%{rvm gemset create "#{gemset_name}"}).chomp
+
+        recognized_ouputs = [
+          /- #gemset created (.*@[^\/]+)$/,
+          /'[^']+' gemset created \((.*)\)\./
+        ]
+
+        recognized_ouputs.each do |format|
+          gem_home = output[format, 1]
+          return gem_home if gem_home
+        end
+
+        raise "I didn't understand rvm's output: #{output}"
+      end
+
+      def current_gem_home
+        rvm_capture('rvm gemset gemdir').chomp
+      end
+
+      def modify_path_string(str)
+        paths = (str || "").split(File::PATH_SEPARATOR)
+        yield(paths).uniq.join(File::PATH_SEPARATOR)
+      end
+
+      def replace_paths(str, old, new)
+        modify_path_string(str) do |paths|
+          paths.delete(old)
+          [new] + paths
+        end
+      end
+
+      def rvm_capture(cmd)
+        rvm_run(cmd)
+        last_command_started.stdout.dup
+      end
+
+      def rvm_run(cmd, options = {})
+        opts = [true, options[:timeout]].compact
+        run_simple(cmd, *opts)
       end
     end
   end
--- a/spec/aruba/api/environment/rvm_spec.rb
+++ b/spec/aruba/api/environment/rvm_spec.rb
@@ -1,0 +1,258 @@
+require 'spec_helper'
+
+RSpec.describe Aruba::Api::Rvm do
+  include_context 'uses aruba API'
+
+  around do |example|
+    Aruba.platform.with_environment do
+      example.run
+    end
+  end
+
+  let(:instance) { @aruba }
+
+  subject { instance }
+
+  describe '.use_clean_gemset' do
+    before do
+      allow(instance).to receive(:set_environment_variable).with(instance, anything)
+    end
+
+    context "when creating gemset was unsuccessful" do
+      before do
+        allow(instance).to receive(:current_gemset_home).with("foobar").and_raise("I didn't understand rvm's output: blah blah blah")
+      end
+
+      it "fails" do
+        expected = "I didn't understand rvm's output: blah blah blah"
+        expect{ subject.use_clean_gemset("foobar") }.to raise_error(expected)
+      end
+    end
+
+    context "when creating gemset succeeded" do
+      let(:gem_path) { "/home/foo/ruby-1.2.3:/home/foo/ruby-1.2.3@global" }
+
+      let(:path) do
+        [
+          "/home/foo/ruby-1.2.3/bin",
+          "/home/foo/ruby-1.2.3@global/bin",
+          "/bar",
+          "/baz"
+        ].join(":")
+      end
+
+      before do
+        allow(instance).to receive(:current_gemset_home).and_return("/home/foo/ruby-1.2.3@foobar")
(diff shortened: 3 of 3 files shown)


maxmeyer:
Thanks a lot. @e2 Is this fix needed soon or can you wait a bit? There are some ongoing changes here #257, here #258 and here #260.

@jarl-dk @mattwynne I like the idea of having a CommandRunner and EnvironmentSetter, but some of those idea are solved more generally in one the PRs given above. Furthermore it's quite a lot of code to solve the "same" problem, isn't it?

I think we also need discuss if we want to keep those gemset methods as part of aruba(-core), as using `bundler` could also solve the same problem. But I must admit I do not use `rvm` myself, but using `bundler` in all my ruby projects. Maybe there are some users who can share their opinions.

maxmeyer:
@limhoff-r7 I think I saw you using the gemset-methods of aruba, is this correct? If yes, why are you using them and not bundler to separate gems?

limhoff-r7:
I was using the gemset methods, but I had to modify them to work.  I eventually had to remove those features because the gemset methods worked locally in OSX, but not on Travis-CI's Ubuntu, which this PR is probably fixing.   I was testing that a command line tool properly added itself to the gemspec, so using bundler alone would have leaked the test environment's gems and not actually tested that bundle install worked in a fresh project.  In general rvm gemsets are a better separation of gems than bundler alone gives.  Too cases that come to mind are yard plugins and some edge cases we've seen when switch back and forth between N and N+1 versions of Rails when doing Rails upgrades from 3.X to 4.X.

e2:
Thanks. There's no rush - I see there's tons of commits in the queue. I'm pretty much finished with this PR, and these extra "mini" classes are there to improve maintainability (edge cases + specs) without messing around with (or relying too much on) anything else in the codebase.

Background: I use only Linux - and I've had my fair share of RubyGems/Bundler/RVM/ZSH related issues. I heavily rely on gemsets (even having things relying on specific versions of Bundler).

Anyway, this should fix a few issues with paths, etc.

maxmeyer:
>  I was testing that a command line tool properly added itself to the gemspec, so using bundler alone would have leaked the test environment's gems and not actually tested that bundle install worked in a fresh project.

That's not clear to me, how can a command line tool add itself to the gemspec? Or do you mean it's found in `PATH`?

@limhoff-r7 @e2 Do you both have some public examples available how you are using the gemset methods?

Thanks.

e2:
TL;DR - the 'use_clean_gemset' method would save lots of development hours if it worked as advertised (and even more so with Docker becoming so popular), and with this PR is becomes possible to safely extend the functionality without breaking things. The alternative is to wrap the tested tool in a shell script - which ideally wouldn't be necessary if this method worked.

@maxmeyer - every time I tried to use it, it never worked properly (even when patching the regexp), so either I ran the tested app through a shell script (to initialize things with rvm directly via a rvm-loaded shell instead of hacking environment variables), or tested it manually (and using aruba only when testing - instead of when developing as well).

Since it never worked (I kept forgetting about that regularly) and there were no specs, I never found enough time to get back to this again. Especially since it always seemed easier to just copy-paste-adapt without needing to patch Aruba itself.

The regexp was broken since 2012: https://github.com/rvm/rvm/commit/8ec6c907c667f8ed9f67ef779e4f6d5348a8a0e2#diff-e10db86b63bfa343e67af19b99514ed9R199

Anyway, `guard` is an example where this functionality would've prevented releasing broken code - or spending lots of time reproducing issues and testing manually, see: https://github.com/guard/guard/tree/master/bin

Guard can be run with and without bundler - and handles edge cases by displaying user-friendly errors. Most often though, users are recommended to run all the tools through bundler exec anyway.

The ability to switch to a gemset and avoid bundler means it's possible to test edge cases of tools (running without bundler) running other tools (using bundle exec).

The alternative to using a gemset would be running a given app in a Docker container - or hacking/debugging custom gemfiles (and polluting the tests with additional steps to create them).

Also, it's easier to detect missing dependencies and missing `require` calls, because it's possible that development gems require files that won't be available during CI runs or on user setups - it's better to catch these before deploying to production. This is especially important for setups using docker - where available gemsets may be very minimalistic.

Even if the use cases seem exotic, there's value when developing/debugging - just like the announcer is valuable, even if there's no project-specific "use case" for it (other than during development). 

It's much easier to debug issues within a minimal sandbox (minimal gemset provided by RVM), than debugging effectively something like `bundle exec cucumber` which would be running something like `HOME=aruba_temp_dir_ BUNDLER_GEMFILE=custom_gemfile bundle exec mytool` (which means running a tool under bundler with a custom gemfile, while the main process with being run with bundler with another gemfile, another $HOME, etc.). `use_clean_gemset` helps debug much faster with less complexity.

Even though this may seem RVM specific, it's an optimization over testing manually inside docker - analogically just like Aruba in-process is an optimization.

Of course the implementation is another story.

mattwynne:
I think this PR looks like it moves the existing code in a positive direction.

On the whole though, I don't like having rvm-specific code in the core of Aruba. It feels wrong to assume that the users will know what RVM is - you could and should be able to use Aruba for any command-line app, whether it's written in Ruby or not.

I think it would be best to extract this into a separate `aruba-contrib` gem eventually, but that can wait.

e2:
@mattwynne - at first I also thought this was RVM specific, but in reality changing `GEM_HOME`, `GEM_PATH` and `PATH` means configuring RubyGems and Bundler, not RVM. 

The method is Ruby-specific. RVM just makes settings things up more convenient than sandboxing things manually for RubyGems/Bundler.

The "only" thing RVM brings here to the table is: creating the directories and symlinking them - `use_clean_gemset` can be implemented without RVM, though it's probably just not worth the effort (just because of the gemset management).

This `use_clean_gemset` implementation actually bypasses most of RVM's functionality, because normally RVM manages the env vars. (RVM here is being used incorrectly because of these variables, but ... that's a different story).

e2:
Can this be merged? (After adapting?). 

RVM support is broken (newer versions) ever since. It's depreciated, but that doesn't mean it shouldn't work. Also, without this patch, there are no RVM specs anyway.

Since Aruba 1.x still hasn't been released for over a year, it doesn't make sense to put this 5 years into the future.

Also, RVM handling is way too tricky to leave it up to users and just expect them to "setup the environment properly". It's almost rocket science (see path handling), so people are going to want this in a plugin, and it doesn't make sense to have an "external aruba plugin" no one will know about.

maxmeyer:
@e2 Yes, it can be merged after adapting. I asked Cucumber-Org-maintainers to create a new repo for `aruba-contrib`. I'm going to give you a ping when the repo has been created. Until this one is created we can go on with this PR.

Some notes toward the code. I'm quite happy, that you take the burden to refactor the code. What I would love to see are some simplifications - e.g. I changed quite a few things in `aruba`'s very own codebase in the area of environment variables.

1. Please don't introduce that many classes to solve the problem
2. Don't pass in `aruba`-context into classes This feels weird
3. Please stick to the approach used earlier and try to use as many methods you find in aruba's code base. There's even a new one called [`#delete_environment_variable`](https://github.com/cucumber/aruba/blob/master/lib/aruba/api/environment.rb#L76) you can use to remove the `env.variable = nil` things.
4. To check if everything is correct you might want to use similar code:
  
  ~~~ruby
  run_simple('rvm create gemset...')
  expect(last_command_started).to have_output('asdf')
  ~~~

5. I like the `GemSet`-thing, but please don't make it just a wrapper and return all information needed to run with normal aruba methods.

~~~ruby
gem_set = GemSet.new(set_name)
run_simple(gem_set.to_command_line)
set_environment_variable('GEM_HOME', gem_set.gem_home_path)
append_environment_variable('PATH', ':' + gem_set.bin_path)

# [...]
~~~

Does this makes sense to you?

e2:
> Please don't introduce that many classes to solve the problem

I removed all the classes. The specs are now hideously interdependent (mocks). Oh well...

> Don't pass in aruba-context into classes. This feels a bit weird.

I wanted to isolate out Aruba internals from getting in the way of tests. And to avoid maintenance problems with all the internal API changes in Aruba. Now the specs are riding off of the Api-extended Runtime instance. 

> Please stick to the approach used earlier and try to use as many methods you find in aruba's code base.

I can't use the prepend/append because RVM uses a special "substitution". 

> There's even a new one called #delete_environment_variable you can use to remove the env.variable = nil things.

Done.

> I like the GemSet-thing, but please make it just a wrapper and return all information needed to run with normal aruba methods.

Replaced with methods.

> Please rebase to master.

Done.

> Could you do me a favor and add some examples in features/steps/rvm.feature and features/api/rvm.feature about how to use the steps / api?

I'd like to, but right now I'm traumatized by Aruba. Maybe if there's a release with a working version first. (I don't want deal with 3 PRs which all depend on each other).

> Since I've never used it, that kind of documentation would be very valuable for other users and even me.

It's just a single method.

 It allows e.g. a gem to "self-test" itself to see if it can be installed. So you can create a gemset, install the gem, do something with it (if it has a bin script) to see if it works after installation, you can check if dependencies can be satisfied, etc. In-process aruba is normally used for this, though it helps to spawn a process in it's own gemset.

In short - it promises full isolation. You can build the gem locally with rake, and then use Aruba to install it into an isolated gemset.

Also, it never worked until now, so I doubt anyone is actually using it. It's self-explanatory for RVM users, anyway. (Just never worked as advertised).

Let me know if there's anything more needed to get this merged.

dg-ratiodata:
> I removed all the classes. The specs are now hideously interdependent (mocks). Oh well...

Yes, I know that. I think the way you build the RVM-class maybe a way we should move `aruba`'s code to, but for now I don't want to introduce another way to structure the code. This can happen in 2.0.0.

For now I moved more and more API tests to the feature files. With that kind of test you don't need to mock anything and get the documentation for free.

> I can't use the prepend/append because RVM uses a special "substitution". 

Oh, dear... :-(

> I'd like to, but right now I'm traumatized by Aruba. 

oh! That would be something I would love to talk about. In what way and why? We can head to https://gitter.im/cucumber/aruba to discuss this if you like.

>It allows e.g. a gem to "self-test" itself to see if it can be installed. So you can create a 
>gemset, install the gem, do something with it (if it has a bin script) to see if it works after 
>installation, you can check if dependencies can be satisfied, etc. In-process aruba is 
>normally used for this, though it helps to spawn a process in it's own gemset.

I think I get it. What I suppose is this:

Move most tests from rspec tests for `use_clean_gemset` to feature tests (`feature/api/use_clean_gemset.feature`. Have a look at other tests to see how they are setup. Doing it this way reduces the problem with mocking. Only leave the tests for edge cases - if there are any in the rspec test suite.

>> I like the GemSet-thing, but please make it just a wrapper and return all information
>> needed to run with normal aruba methods.

> Replaced with methods.

As said before I think the GemSet.new-class really makes sense here. Please move all methods from https://github.com/cucumber/aruba/pull/261/files#diff-846e84d232a830030fcf744d12dcafaaR48 and all other path methods like `current_gem_home` into `GemSet`-class. We need to get rid off of all those private methods. Without that we're back into trouble because users might find those small methods helpful and start using them although they're not part of the official API.

maxmeyer:
@e2 @mattwynne This functionality will be removed in `aruba`. Please head over to https://github.com/cucumber/aruba-contrib and open a PR there. Thanks a lot.

e2:
Seeing this: https://github.com/cucumber/aruba-contrib/blob/master/lib/aruba/contrib/rvm.rb I give up.

I just don't feel my time is respected. I've already spent too much time redoing this over so many times. 

And if I have to wait weeks or months just to get a PR reviewed ... just to have it rejected because of "too many classes" or something like that ... it's just a massive waste of life.

Feel free to copy/adapt my PR. I'm not interested in preparing and testing it all over again - with no ETA or timeframe. I'm just not that bored, or desperate, sorry.

I know no one is getting paid for review it, but I'm not getting paid for redoing it over and over again - and still not being able to expect it to be released within a sane amount of time. I created this PR almost a year ago. 

Initially I wanted to just make it work. Now I'm like a maintainer without access to get it merged, released, improved and re-released. I wish I just forked it and released it on my own. Waiting for a doctor takes less time...

So if anyone wants to pick this up ... go ahead.

e2:
FWIW, I created https://github.com/e2/gem_isolator#usage

It isn't RVM specific and it uses bundler for setting up the sandbox.

maxmeyer:
Thanks for sharing.

PR #275, opened 2015-07-02
Updating Vendored JQuery

DVG:
The currently vendored version of JQuery has an open cross-site scripting vunerability (CVE-2011-4969). While this is probably not even a thing to most people, in an enterprise risk-averse environment a project using aruba can potentially be blocked from deploying due to the vendored version of JQuery even being present. Updating to the latest JQuery.


mattwynne:
Thanks but now I'm wondering why on earth we have jQuery as part of this library?! Is this cucumber-reports again?

DVG:
Yes, it's just to enable a click event on a generated report.

maxmeyer:
@DVG What are you using `aruba-reports` for? For debugging? For generating documentation?

DVG:
Neither, actually, the vendored version of Jquery just threw up a flag in a project using arubu because of the above mentioned CVE. I'm not using the reporting aspect at all, but it is causing some issues with internal compliance rules.

maxmeyer:
So it would help you if we remove it?

DVG:
Thanks!

maxmeyer:
I pushed 0.7.0 to rubygems. Please let me know if you have any trouble with this release because it changed a lot under the hood.

DVG:
Nice! Thanks

Issue #285, opened 2015-07-14
File permissions are platform specific; tests fail on Windows


weedySeaDragon:
For both 0.7.4 and 0.8.0 the tests for file permissions in <code>spec/aruba/api_spec.rb</code> are using Unix-platform specific tests. Thus these tests will always fail on Windows platforms.  
From Ruby Core 2.2.2 - File (http://ruby-doc.org/core-2.2.2/File.html) :  

```
In the description of File methods, permission bits are a **platform-specific**  
set of bits that indicate permissions of a file.  

[... paragraph on Unix permission bits...]

On non-Posix operating systems, there may be only the ability to make a file read-only  
 or read-write.  In this case, the remaining permission bits will be synthesized to resemble  
 typical values. For instance, on Windows NT the default permission bits are 0644,   
 which means read/write for  owner, read-only for all others. The only change that can   
be made is to make the file read-only, which is reported as 0444.
```

dg-ratiodata:
Ok. Reading this, it is enough to change the permissions used within the tests to make them pass on windows? Are you willing to send a PR for this as well?

weedySeaDragon:
Yup, I can do a PR.  I'll need help understanding the intent of some of the tests, probably.  The biggest issue will be that Windows just doesn't support the variety of permissions / level of granularity that *nix systems do.  I'll start working and will ask my questions here.

maxmeyer:
Perfect. Maybe it's best you add some feature tests first.

```
mkdir -p features/api/filesystem/
touch features/api/filesystem/which.feature
```

Have a look at `features/api/filesystem/which.feature` for some examples on how to write the tests.

weedySeaDragon:
OK -- I'll be sure to do the features first and have you give them a once-over to be sure I've got it right.

dg-ratiodata:
Perfect.

weedySeaDragon:
So your idea was spot on, @maxmeyer : the tests pass if the file permissions used in the tests are changed to ones that Windows 'understands'. Specifically, changing the tests to use 0644 (or '0644')  will make them all pass.
But if it's important that those particular file permissions are used, then we'll need different tests for Windows.

Issue #284, opened 2015-07-14
v 0.8.0 'which' is unix specific (doesn't work on Windows)


weedySeaDragon:
'which' is a unix-specific command; it doesn't exist on Windows  ('where' is a close equivalent).  Also, checking a PATH for the existence of a command does not work the same in Windows.

There is code in <code>spawn_process.rb</code> and <code>debug_process.rb</code> that uses 'which' and checks the PATH.  This won't work on Windows.  I'm not clear on what the intent is.  I can help with a PR if you can help me better understand the intent.

maxmeyer:
Somehow path resolution fails for jruby + childprocess. That's why I added which.

weedySeaDragon:
Ok, but now you've broken Windows. :-0
If that approach (checking the PATH) makes sense for the unix systems, then sounds like a conditional is needed so that you have a working alternative approach for windows.  Since you'll need that conditional check in a few places (e.g. 'is this running on windows?') I suggest that you add a utility method somewhere for that check.  ChildProcess has a good set of methods for checking the os.  I'm not sure if you can use them; I don't know if you have a ChildProcess instantiated everywhere when you need to check the os.  Or you can use the same approach that ChildProcess uses.
Or you can certainly make some child classes of SpawnProcess that each know how to build and run a process for their respective OS. (or OS-specific classes that are components of SpawnProcess, etc.) I don't know if that amount of work is worth it right now for you, and I don't know how much work that might or might not be.  Just throwing it out there.

I can certainly help write or incorporate the methods for checking which os it is.

dg-ratiodata:
BTW: I'm not using the `which`-utility but a helper method https://github.com/cucumber/aruba/blob/master/lib/aruba/platform.rb#L177. First of all I like to  get that one fixed. If you can help with that?

weedySeaDragon:
Thanks for straightening me out on that, @dg-ratiodata .  My bad for assuming it was the <code>which</code> utility.  Since it's an aruba method, that should make it easier to address.  
Yes, I can help.  
[...]

Issue #283, opened 2015-07-14
v 0.7.4 ChildProcess on Windows requires "cmd.exe /c"


weedySeaDragon:
Using version 0.7.4 on Windows, in <code>run! Child.Process.build(....) </code> fails with a <code>ChildeProcess LaunchError</code>.  Here's a simple example running a cucumber scenario that uses aruba to call "rake spec"  
Here's the cucumber scenario:  
```
Feature: Newly created rails gem is A-OK

  A rails gem that has just been created by the bundler should pass basic spec tests

  @announce
  @rail_gem
  Scenario: A newly created gem passes all spec tests
    Given the empty file named "test.tmp"
    When I run `rake spec`
    Then the output should contain "0 failures"
```
and here's the error that happens when running it:
```
> bundle exec rake cucumber

Feature: Newly created rails gem is A-OK
  
  A rails gem that has just been created by the bundler should pass basic spec tests

  @announce @rail_gem
  Scenario: A newly created gem passes all spec tests # features\generator_new_gem.feature:7
    Given the empty file named "test.tmp"             # aruba-0.7.4/lib/aruba/cucumber.rb:47
$ cd C:/Users/ashley/RubymineProjects/ammeter
$ rake spec
    When I run `rake spec`                            # aruba-0.7.4/lib/aruba/cucumber.rb:119
      The system cannot find the file specified. (error #2 -(size  44) (Aruba::LaunchError)
      features\generator_new_gem.feature:9:in `When I run `rake spec`'
    Then the output should contain "0 failures"       # aruba-0.7.4/lib/aruba/cucumber.rb:166

Failing Scenarios:
cucumber features\generator_new_gem.feature:7 # Scenario: A newly created gem passes all spec tests

1 scenario (1 failed)
3 steps (1 failed, 1 skipped, 1 passed)
0m0.047s
```

Per https://github.com/jarib/childprocess/issues/59, on Windows, ChildProcess requires you to start any shell commands with "cmd.exe /c".  The following code works for me:

```
def run!
        @process = ChildProcess.build(*Shellwords.split(@cmd))

        # Windows needs to prepend commands with 'cmd.exe /c' @see from https://github.com/jarib/childprocess/issues/59
        @process = ChildProcess.build(*(Shellwords.split(@cmd).unshift("cmd.exe", "/c"))) if ChildProcess.windows?

```

Would you like a PR for this?

dg-ratiodata:
[...]

PR #290, opened 2015-07-15
provide methods to query which OS is running. Use FFI::Platform methods

weedySeaDragon:
I've added some methods into (lib)/api/platform.rb to query which OS is running. Since FFI is already being included by the project and FFI already has excellent methods for querying the OS, the methods are essentially pass-thru's to what FFI does.

If it's somehow not good to use ffi (it did require a require 'ffi' in platform.rb) I can use the already required rbconfig. (Tho I'll likely be re-writing what FFI has done.)


weedySeaDragon:
I've now added **RSpec methods so that we can test -- and change -- the current operating system, ruby_platform, and ruby_engine.** The current os is gotten from FFI::Platform; these test mock FFI::Platform with a class_double.  These spec also use stub_constant for RUBY_ENGINE and RUBY_PLATFORM -- so those can be changed, too.  
The intent is that these tests are the basis for being able to instrument those values (the current OS, RUBY_PLATFORM, and RUBY_ENGINE) as needed so we can test the system behavior without always having to run a travis or appveyor job.  This should make it easier to develop tests for behavior based on the state of one of those.  (Like  #283, #284, #285)  
The tests are in a file named **<code>spec/aruba/platform/os_platforms_engines_spec.rb
</code>** -- which I think is a poor file name.  It's descriptive, but seems long.  Should the file name be changed?  I want to be sure to keep with the standards and feel of this project.  
What do you think of these, @dg-ratiodata  (and @maxmeyer )?  I want to get these right so that I can build on them for the other windows-related PRs that I'm working on.

mattwynne:
FYI @weedySeaDragon, @dg-ratiodata and @maxmeyer are the same person. It confuses me too.

weedySeaDragon:
@mattwynne  ya, I learned that the other night on Gitter. I referenced both of those as kind of a joke about exactly that. :-)

mattwynne:
Maybe I'm missing something but I don't see why we need the specs in `os_platforms_engines_spec.rb` at all. Are they designed to be used elsewhere, or are they just to test the new methods added to `Aruba::Platform` in this PR?

weedySeaDragon:
Yes and yes.  These do test the new methods in Aruba::Platform, but then can be expanded (perhaps as shared examples or contexts)  to also be used for the PRs that require different behavior based on the current OS.  
We'll need a way to test the behaviors that are different on Windows, like the file permissions (#285), the different spawned process (#283), etc.  In order to do that kind of testing, we first need to instrument the current test configuration (context, literally) so that the answer to "on we running on windows?" is controlled.  These specs provide that.   I want to make sure these are OK before I go much farther with these.  
If using these as the basis for some of the testing that needs to be done with the other os-specific PRs/issues, is **not** the way to go, then now's the time for me to understand that.  Make sense?

mattwynne:
Rather than spreading the conditional logic about the OS around the code, I'd prefer to isolate it at the edges. Let's build some different instances of the `Platform` duck-type for the different platforms, and then we can keep all the OS-specific code in one place.

I'll see if I can knock up a PR to show you what I mean.

weedySeaDragon:
Sounds good.  closing this.

maxmeyer:
@mattwynne I started breaking up `Platform` and even  `Api#methods` into smaller classes. Just for you to know.

weedySeaDragon:
Should've just reverted the commits (not closed this whole thing)

(max - sounds like you and should get coordinated about what should be done on Platform and SpawnProcess so that we don't duplicate work.)

maxmeyer:
> (max - sounds like you and should get coordinated about what should be done on Platform and SpawnProcess so that we don't duplicate work.)

Yes. Personally I like to see what @mattwynne has in mind with ` I'd prefer to isolate it at the edges. `.

From my point of view, having a class for edge cases (WindowsEnvironment, PosixEnvironment, WindowsWhich, PosixWhich) would be the way to go as this approach can handle both different data structures and different algorithms.

I don't want to bloat `Platform` too much as it includes way to many methods already.

mattwynne:
When I say "edges" I mean the edge of the architecture - use a [ports & adapters](http://alistair.cockburn.us/Hexagonal+architecture) design so we can pass in an adapter for the platform-specific stuff, and all method calls that need platform-specific implementations go across a common port. I think all of us are on the same page now (@weedySeaDragon and I already caught up about this on gitter)

dg-ratiodata:
See #299 for a possible implementation for this.

dg-ratiodata:
As we've got the check in `Platform` now it does not make sense to add those methods to `Platform`. Sorry @weedySeaDragon for the time you spent on this. I used your feedback and changed the Unix.match?-method a bit.

PR #293, opened 2015-07-19
Use the right Environment class based on the OS

weedySeaDragon:
This depends on PR #290  because we need to know which OS is running in order to pick the correct Environment class

### DIFF
--- a/lib/aruba/environment.rb
+++ b/lib/aruba/environment.rb
@@ -1,121 +1,14 @@
+require 'aruba/environments/unix_environment'
+require 'aruba/environments/windows_environment'
+
 module Aruba
-  class Environment
-    private
-
-    attr_reader :env
-
-    public
-
+  class Environment < SimpleDelegator
     def initialize
-      @env = ENV.to_hash
-    end
+      environments = []
+      environments << Aruba::Environments::WindowsEnvironment.new
+      environments << Aruba::Environments::UnixEnvironment.new
 
-    # Update environment with other en
-    #
-    # @param [#to_hash, #to_h] other_env
-    #   Another environment object or hash
-    #
-    # @yield
-    #   Pass block to env
-    def update(other_env, &block)
-      if RUBY_VERSION <= '1.9.3'
-        # rubocop:disable Style/EachWithObject
-        other_env = other_env.to_hash.inject({}) { |a, (k, v)| a[k] = v.to_s; a }
-        # rubocop:enable Style/EachWithObject
-      else
-        other_env = other_env.to_h.each_with_object({}) { |(k, v), a| a[k] = v.to_s }
-      end
-
-      env.update(other_env, &block)
-
-      self
-    end
-
-    # Fetch variable from environment
-    #
-    # @param [#to_s] name
-    #   The name of the variable
-    #
-    # @param [Object] default
-    #   The default value used, if the variable is not defined
-    def fetch(name, default)
-      env.fetch name.to_s, default
-    end
-
-    # Check if variable exist
-    #
-    # @param [#to_s] name
-    #   The name of the variable
-    def key?(name)
-      env.key? name.to_s
-    end
-
-    # Get value of variable
-    #
-    # @param [#to_s] name
-    #   The name of the variable
-    def [](name)
-      env[name.to_s]
-    end
-
-    # Set value of variable
-    #
-    # @param [#to_s] name
-    #   The name of the variable
-    #
-    # @param [#to_s] value
-    #   The value of the variable
-    def []=(name, value)
-      env[name.to_s] = value.to_s
-
-      self
-    end
-
-    # Append value to variable
-    #
-    # @param [#to_s] name
-    #   The name of the variable
-    #
-    # @param [#to_s] value
-    #   The value of the variable
-    def append(name, value)
-      name = name.to_s
-      env[name] = env[name].to_s + value.to_s
-
-      self
-    end
-
-    # Prepend value to variable
-    #
-    # @param [#to_s] name
-    #   The name of the variable
-    #
-    # @param [#to_s] value
-    #   The value of the variable
-    def prepend(name, value)
-      name = name.to_s
-      env[name] = value.to_s + env[name].to_s
-
-      self
-    end
-
-    # Convert to hash
-    #
-    # @return [Hash]
-    #   A new hash from environment
-    def to_h
-      if RUBY_VERSION < '2.0'
-        env.to_hash.dup
-      else
-        env.to_h.dup
-      end
-    end
-
-    # Reset environment
-    def clear
-      env.clear
-
-      self
+      super environments.find(&:match?)
     end
   end
 end
--- a/lib/aruba/environments/unix_environment.rb
+++ b/lib/aruba/environments/unix_environment.rb
@@ -1,0 +1,11 @@
+require 'aruba/environments/basic_environment'
+
+module Aruba
+  module Environments
+    class UnixEnvironment < BasicEnvironment
+      def match?(*)
+        Aruba::Platform.unix?
+      end
+    end
+  end
+end
--- a/lib/aruba/platform.rb
+++ b/lib/aruba/platform.rb
@@ -1,5 +1,6 @@
 require 'rbconfig'
 require 'pathname'
+require 'ffi'
 
 require 'aruba/platform/simple_table'
 
@@ -182,19 +183,16 @@ module Aruba
     # rubocop:disable Metrics/MethodLength
     # rubocop:disable Metrics/CyclomaticComplexity
     def which(program, path = ENV['PATH'])
-      on_windows = false
-      on_windows = true if File::ALT_SEPARATOR
-
       program = program.to_s
 
-      path_exts = ENV['PATHEXT'] ? ('.{' + ENV['PATHEXT'].tr(';', ',').tr('.','') + '}').downcase : '.{exe,com,bat}' if on_windows
+      path_exts = ENV['PATHEXT'] ? ('.{' + ENV['PATHEXT'].tr(';', ',').tr('.','') + '}').downcase : '.{exe,com,bat}' if windows?
 
       raise ArgumentError, "ENV['PATH'] cannot be empty" if path.nil? || path.empty?
 
       # Bail out early if an absolute path is provided or the command path is relative
       # Examples: /usr/bin/command or bin/command.sh
       if Aruba::Platform.absolute_path?(program) || Aruba::Platform.relative_command?(program)
-        program += path_exts if on_windows && File.extname(program).empty?
+        program += path_exts if windows? && File.extname(program).empty?
 
         found = Dir[program].first
 
@@ -211,7 +209,7 @@ module Aruba
 
         # Dir[] doesn't handle backslashes properly, so convert them. Also, if
         # the program name doesn't have an extension, try them all.
-        if on_windows
+        if windows?
           file = file.tr("\\", "/")
           file += path_exts if File.extname(program).empty?
         end
@@ -220,7 +218,7 @@ module Aruba
 
         # Convert all forward slashes to backslashes if supported
         if found && Aruba::Platform.executable_file?(found)
-          found.tr!(File::SEPARATOR, File::ALT_SEPARATOR) if on_windows
+          found.tr!(File::SEPARATOR, File::ALT_SEPARATOR) if windows?
           return found
         end
       end
@@ -229,6 +227,16 @@ module Aruba
     end
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/CyclomaticComplexity
+
+    # return true if the OS running is a Unix system according to FFI
+    def unix?
+      FFI::Platform.unix?
+    end
+
+    # return true if the OS running is Windows according to FFI
+    def windows?
+      FFI::Platform.windows?
+    end
 
     module_function :detect_ruby, \
       :current_ruby, \
@@ -252,6 +260,8 @@ module Aruba
       :getwd, \
       :which, \
       :simple_table, \
-      :write_file
+      :write_file, \
+      :unix?, \
+      :windows?
   end
 end
--- a/spec/aruba/api_spec.rb
+++ b/spec/aruba/api_spec.rb
@@ -698,7 +698,7 @@ describe Aruba::Api  do
       context 'when file_name contains fixtures "%" string' do
         let(:runtime) { instance_double('Aruba::Runtime') }
         let(:config) { double('Aruba::Config') }
-        let(:environment) { instance_double('Aruba::Environment') }
+        let(:environment) { object_double(Aruba::Environment.new) }
 
         let(:klass) do
           Class.new do
--- a/spec/aruba/environment_spec.rb
+++ b/spec/aruba/environment_spec.rb
@@ -1,0 +1,10 @@
+require 'aruba/environment'
+require 'spec_helper'
+
+describe Aruba::Environment do
+  subject(:env) { described_class.new  }
+
+  it { expect(env['PATH']).not_to be_nil}
+
+  it { expect(env.clear.to_h).to be_empty}
+end
--- a/lib/aruba/environments/basic_environment.rb
+++ b/lib/aruba/environments/basic_environment.rb
@@ -1,0 +1,131 @@
+module Aruba
+  module Environments
+    class BasicEnvironment
+      private
+
+      attr_reader :env
+
+      public
+
+      def initialize
+        @env = ENV.to_hash
+      end
+
+      # Update environment with other en
+      #
+      # @param [#to_hash, #to_h] other_env
+      #   Another environment object or hash
+      #
+      # @yield
+      #   Pass block to env
+      def update(other_env, &block)
+        if RUBY_VERSION <= '1.9.3'
+          # rubocop:disable Style/EachWithObject
+          other_env = other_env.to_hash.inject({}) { |a, (k, v)| a[k] = v.to_s; a }
+          # rubocop:enable Style/EachWithObject
+        else
+          other_env = other_env.to_h.each_with_object({}) { |(k, v), a| a[k] = v.to_s }
+        end
+
+        env.update(other_env, &block)
+
+        self
+      end
+
+      # Fetch variable from environment
+      #
+      # @param [#to_s] name
+      #   The name of the variable
+      #
+      # @param [Object] default
+      #   The default value used, if the variable is not defined
+      def fetch(name, default)
+        env.fetch name.to_s, default
+      end
+
+      # Check if variable exist
+      #
+      # @param [#to_s] name
+      #   The name of the variable
+      def key?(name)
+        env.key? name.to_s
+      end
+
+      # Get value of variable
+      #
+      # @param [#to_s] name
+      #   The name of the variable
+      def [](name)
+        env[name.to_s]
+      end
+
+      # Set value of variable
+      #
+      # @param [#to_s] name
+      #   The name of the variable
+      #
+      # @param [#to_s] value
+      #   The value of the variable
+      def []=(name, value)
+        env[name.to_s] = value.to_s
+
+        self
+      end
+
+      # Append value to variable
+      #
+      # @param [#to_s] name
+      #   The name of the variable
+      #
+      # @param [#to_s] value
+      #   The value of the variable
+      def append(name, value)
+        name = name.to_s
+        env[name] = env[name].to_s + value.to_s
+
+        self
+      end
+
+      # Prepend value to variable
+      #
+      # @param [#to_s] name
+      #   The name of the variable
+      #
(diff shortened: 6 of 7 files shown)


maxmeyer:
Can you please rebase?

weedySeaDragon:
Dang.  You bet.  Sorry about that.

maxmeyer:
Use something better, like Arch Linux. ;-)

weedySeaDragon:
if you want to be my sysadmin on call 24/7, I'll make the change. ;-)  (I will switch over, but I don't have the bandwidth right now.  In a few months, I hope.)

maxmeyer:
Not sure what you did, but there are 57 commits in this PR. Please make sure, that only your changes are part of the commit. Maybe you need to do something like that: `git rebase origin/master`

weedySeaDragon:
I think it's fixed now.  I'm showing a normal amount of commits -- and they're just my stuff (as it should be).  
I was using the 'GitHub for Windows' app and it was pulling in commits.  (It's trying to make things 'easy' I'm sure, but instead it gives you no way to really have control over configuration, etc. sigh.  when I did a merge, I didn't realize it was trying to keep this branch in sync with upstream.  hence it pulled in a lot.  then gave me no way to undo.  Back to go old command line git for me.)

spec/aruba/environment_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'aruba/environment'
+require 'spec_helper'
+
+describe Aruba::Environment do
+  let(:environ) { Aruba::Environment.new  }
```
maxmeyer: Just make it env here, please.

spec/aruba/environment_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  it 'should find PATH' do
+    expect(environ['PATH']).not_to be_nil
+  end
+
+  it '#clear' do
```
maxmeyer: it -> describe. Why is used `puts` here? 

`it { expect(environ.clear.to_h).to be_empty }`

spec/aruba/environment_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'aruba/environment'
+require 'spec_helper'
+
+describe Aruba::Environment do
+  let(:environ) { Aruba::Environment.new  }
+
+  it 'should find PATH' do
```
maxmeyer: just `it {  expect ... }`

lib/aruba/platform.rb
```diff
@@ -230,6 +231,31 @@ def which(program, path = ENV['PATH'])
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/CyclomaticComplexity
 
+    # return true if the OS running is Windows according to FFI
+    def on_windows?
```
maxmeyer: I thougt abou this a bit. Please remove all `on_*` from the methods.

lib/aruba/environment/windows_environment.rb
```diff
@@ -0,0 +1,11 @@
+require 'aruba/environment/basic_environment'
+
+module Aruba
+  class WindowsEnvironment < BasicEnvironment
+    def initialize
```
maxmeyer: What about something like this

```ruby
ENV.to_hash.inject({}) { |a, (k,v)| a[k.to_s.uppercase] = v }
```

If windows is case-insensitive for ENV, can we savely assume, that the code works for a program written in C, C# or whatever?

lib/aruba/environment/basic_environment.rb
```diff
@@ -0,0 +1,129 @@
+module Aruba
```
maxmeyer: Please move to `environments` -> with `s`.

lib/aruba/environment.rb
```diff
@@ -1,121 +1,12 @@
...
-        env.to_h.dup
-      end
-    end
-
-    # Reset environment
-    def clear
-      env.clear
-
-      self
+      environ = Aruba::Platform.on_windows? ? Aruba::WindowsEnvironment : Aruba::UnixEnvironment
```
maxmeyer: No, if please. 

Add something like this to the class

```ruby
def initialize
  @environments = []
  @environments << Environments::WindowsEnvironment
  @environments << Environments::UnixEnvironment

  super @environments.find { |e| e.match? }
end
```

Add the following to the other classes

```ruby
class WindowsEnvironment < BasicEnvironment
  def self.match?(*)
    Aruba::Platform.windows?
  end

  def initialize
    # [..]
  end
end
```

And for the other one

```ruby
class PosixEnvironment < BasicEnvironment
  def self.match?(*)
     true
  end
end

File: lib/aruba/environment.rb [hunk 8]
```diff
@@ -1,121 +1,12 @@
-module Aruba
-  class Environment
-    private
-
-    attr_reader :env
-
-    public
+require 'aruba/environment/unix_environment'
+require 'aruba/environment/windows_environment'
+require 'ffi'
```
maxmeyer: No needed in this file? Correct? Please add only/everywhere where required.

lib/aruba/environment/unix_environment.rb
```diff
@@ -0,0 +1,6 @@
+require 'aruba/environment/basic_environment'
+
+module Aruba
+  class UnixEnvironment < BasicEnvironment
```
maxmeyer: Posix instead of Unix, please.

maxmeyer:
Ok. @mattwynne Does this comply with the things you have in mind?

@weedySeaDragon If we're finished with review, can you please squash to one / two commits, containing the changed functionality and the tests.

lib/aruba/environment/windows_environment.rb
weedySeaDragon: I did think about that (just changing every ENV name (=key) to uppercase.  But I didn't know if  changes to ENV  would be 'rolled back' after the program ends. So I only wanted to make minimal changes to ENV.  Even though Windows doesn't recognize case, perhaps someone is counting on one of their ENVs being mixed (or whatever) case and is accessing the info some other way. (Through the registry, perhaps.)   
As I started working on a '<code>which</code>' for Windows, I created a method for WindowsEnvironments: given an ENV name, convert it to upper case and save it. Thus we can change (<code>.upcase</code>) only the ENV vars that we really need to deal with.
Whatever way you think is best, I'll make it so. :-)   

> If windows is case-insensitive for ENV, can we savely assume, that the code works for a program written in C, C# or whatever?  

I don't quite understand your question. Which *code* are you referring to?  
Are you asking 'is this the approach that all programs must take when they're dealing with Windows ENVs?'  the answer is yes. (But I don't think that's what you're asking.)

lib/aruba/environment/basic_environment.rb
weedySeaDragon: So WindowsEnvironment -> WindowsEnvironments
 PosixEnvironment -> PosixEnvironments
BasicEnvironment -> BasicEnvironments  
what about Aruba::Environment ? also make plural?   (Doing so will mean touching code elsewhere, of course)

maxmeyer: Sorry, for not being more clear:

PosixEnvironment -> Environments::PosixEnvironment
BasicEnvironment -> Environments::BasicEnvironment

lib/aruba/environment/windows_environment.rb
maxmeyer: > ut I didn't know if changes to ENV would be 'rolled back' after the program ends. 

They are local to that sub-process. So no problem here. Let's try my version first. If someone complains, we will change that.

> Are you asking 'is this the approach that all programs must take when they're dealing with Windows ENVs?' the answer is yes. (But I don't think that's what you're asking.)

Yes, this was my question. :-) Should work on my English. Just try my suggestion first.

weedySeaDragon: Sounds good.  (And your English really, really good.  I think most of my questions are because I just don't know the way you think -- no matter which language. :-) )

lib/aruba/environment.rb
weedySeaDragon: OK.  Except that we still need the <code>posix?</code>  (which was: <code>on_unix?</code> ) and need to use that in PosixEnvironments match:
```
class PosixEnvironments < BasicEnvironments
  def self.match?(*)
     Aruba::Platform.posix?
  end
end
```
otherwise if you're on Windows, then both of WindowsEnvironments and PosixEnvironments return <code>true</code>.

lib/aruba/environment/unix_environment.rb
weedySeaDragon: I don't think it's a good idea to use *Posix* instead of Unix:
- POSIX is a set of compliance standards that applies to many operating systems.  It's not == Unixe or Linux. If an OS meets the Posix standard, it's "Posix compliant."  Many flavors of *nix systems are Posix compliant. I know that Posix is often used as a shorthand for a group of *nix systems, but that's using the term pretty loosely. (And incorrectly)  It's important here to be more accurate.
- We don't have a way to test to see if a system *"is posix?"*  (Which really means "is POSIX compliant.")  FFI doesn't have a test for it.(We _could_ implement a valid test. we'd have to get the list of Posix-compliant systems and ensure that info is kept updated.) 
 Someone reading through the code would see <code>posix?</code> and assume that we're using a valid test for Posix. -- but we're not.    Even if we just use the test <code> !windows? </code> we still need to name the class.

So I would argue that using Posix isn't really the way to go here. 
I hope I'm not sounding argumentative -- I don't mean to.  Just want to make sure everything is considered.

lib/aruba/environment.rb
maxmeyer: Just check for unix and name it UnixEnvironment, with out "s" at the end.

lib/aruba/environment/unix_environment.rb
maxmeyer: Unix is ok for me, as long as Linux is considered to be a "unix".

weedySeaDragon:
OK -- see how that all looks.  I've made the changes and squashed.  MRI v 1.9.2 is failing in Travis but I'm not sure why.  (I don't have 1.9.2 installed yet.  Will do it later todayish.)  
Thanks for your patience in getting me up to speed on the style and expectations.  It should get easier. :-)

maxmeyer:
No, problem. More help, more fun. Thanks for your feedback and your time you spend for aruba. :-)

lib/aruba/environment.rb
```diff
@@ -1,121 +1,14 @@
...
-
-    # Reset environment
-    def clear
-      env.clear
+      environments = []
+      environments << Aruba::Environments::WindowsEnvironment.new
+      environments << Aruba::Environments::UnixEnvironment.new
 
-      self
+      super environments.find(&:match?)
```
mattwynne: \o/ @skmetz would be proud!

spec/aruba/api_spec.rb
```diff
@@ -698,7 +698,7 @@
       context 'when file_name contains fixtures "%" string' do
         let(:runtime) { instance_double('Aruba::Runtime') }
         let(:config) { double('Aruba::Config') }
-        let(:environment) { instance_double('Aruba::Environment') }
+        let(:environment) { object_double(Aruba::Environment.new) }
```
mattwynne: why `object_double` when we just used `double` on the line above. Let's be consistent.

mattwynne:
This looks like a step in the right direction. I have a couple of comments:

- At first, I was confused by the names. The existing use of the term _environment_ is really just to _model environment variables_, but it took me a while to see that - perhaps because of context of recent conversations about using polymorphism for platforms. Maybe `WindowsEnvironmentVariables` etc would be a better set of names just to make this really clear.
- eventually, I want all those `windows?` style predicates to disappear from the code. I will tolerate them for now though :)

lib/aruba/environments/unix_environment.rb
```diff
@@ -0,0 +1,11 @@
+require 'aruba/environments/basic_environment'
+
+module Aruba
+  module Environments
+    class UnixEnvironment < BasicEnvironment
+      def match?(*)
+        Aruba::Platform.unix?
```
dg-ratiodata: Please `true` here. This will be the `"default" and should always matches.

lib/aruba/platform.rb
```diff
@@ -182,19 +183,16 @@ def simple_table(hash)
     # rubocop:disable Metrics/MethodLength
     # rubocop:disable Metrics/CyclomaticComplexity
     def which(program, path = ENV['PATH'])
-      on_windows = false
-      on_windows = true if File::ALT_SEPARATOR
-
       program = program.to_s
```
dg-ratiodata: Let's move this into separate classes. And use something like `Platform::WhichCommand.new(path).which(program)`. Similar to `EnvironmentVariables` as @mattwynne suggested.

```ruby
def which(program, path = ENV['PATH'])
  Platform::WhichCommand.new(path).which(program)
end
```

I'm not sure where to move the created files best, maybe under `lib/aruba/platform/which_command` and then create another file `lib/aruba/platform/which_command.rb`? Maybe we should also move the `environment.rb`-files as well.

Then create some classes:

```ruby
module Aruba
  class BasicWhichCommand
    # common code
  end
```
```ruby
  class UnixWhichCommand < BasicWhichCommand
    private

    attr_reader :path

    public
    def initialize(path)
      @path = path
    end

    def which(program)
    end

    def self.match?(*)
      true
    end
  end
```
```ruby
  class WindowsWhichCommand < BasicWhichCommand
    private

    attr_reader :path

    public
    def initialize(path)
      @path = path
    end

    def which(program)
    end

    def self.match?(*)
      Platform.windows?
    end
  end
```
```ruby
  class WhichCommand
    def initialize
      @whiches = []
      @whiches = WindowsWhichCommand
      @whiches = UnixWhichCommand

    end

    def which(program, path = ENV['PATH'])
      @which.find { |w| w.match? }.new(path).which(program)
    end
  end
end
```

lib/aruba/platform.rb
```diff
@@ -230,6 +228,16 @@ def which(program, path = ENV['PATH'])
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/CyclomaticComplexity
 
+    # return true if the OS running is a Unix system according to FFI
+    def unix?
+      FFI::Platform.unix?
+    end
```
dg-ratiodata: Unix will be default. So we do not need this method. All class handling unix will return true on `match?`. Otherwise we would need to check on `solaris?`, `mac?` etc. as well.

lib/aruba/platform.rb
mattwynne: Nice idea to have lots of little objects. Doesn't all need to go into one PR though.

mattwynne: Also, no need to duplicate the match logic. We can have Platform be the factory for everything.

spec/aruba/api_spec.rb
weedySeaDragon: You're absolutely right, @mattwynne .  Except in this situation, ```instance_double``` would no longer work on ```Aruba::Environment``` because it's now a ```SimpleDelegator``` and as such, Rspec won't verify the ```Aruba::Environment``` double. (Using ```instance_double```, RSpec will report that 'environment (Aruba::Environment.new) can't respond to *clear*' and other methods that are delegated to the ```*Environment``` classes.  )  
If there is a way that RSpec can verify it, I can turn it back in to an ```instance_double```  
Should there be a short comment at the end of the new line 701 explaining this? Something like ```# object_double required because Aruba::Environment is a SimpleDelegator``` ?

lib/aruba/environments/unix_environment.rb
weedySeaDragon: Making that change doesn't work.  With that change in place, when running on Windows, both ```UnixEnvironment.match``` and ```WindowsEnviornment.match``` return ```true``` which leads to the wrong ```Environment`` being created with ``` super environments.find(&:match?)```.
Only *one* of the classes should return true at any given time.  (I tried it on my system to be sure before committing.)  
Unless I'm missing something about this.

lib/aruba/platform.rb
weedySeaDragon: I can see taking this out, since the classes are set up such that the ```Basic*``` class has the default behavior and that behavior is for unix systems.  Thus we only care about *not unix* systems, which == *windows*.  (And so the ```Windows*``` child classes have the variant behavior).    The value of leaving it in is so that the intentions are explicit to anyone reading the code.  It's clear not only that we care about Windows (because we have to do something different), but also that the default behavior is for *unix* systems.  In this case, since the class names are ```Unix*``` I'd say that the intention is already clear enough.
(But if it's left in, or just for future reference:  ```FFI::Platform.unix?``` does return true for all of those (```mac?``` ```solaris?``` etc.). It actually just returns ```!windows?``` -- which is exactly our case.)

lib/aruba/platform.rb
weedySeaDragon: @dg-ratiodata If there still needs to be  ```which```  functionality, then it should be renamed to something like  ```command_valid?``` because ```which``` is a unix-specific term (command).    That is what is really happening in that method:  doing whatever is necessary to check that the program is valid so that we can send it along to ChildProcess and know it will work.  Using that name --```command_valid?```  or some variation -- means that we can do all of the checks that are necessary -- not just checking if it's on the path. We also have to ensure it's executable, for example.  This allows platform-specific checks as well as platform-specific implementations of the checks.
Someone from a Windows point of view reading the code won't understand why the method is called ```which```.  There is no ```which``` on Windows; that is meaningless to a Windows person.  On Windows there's a ```where``` command to accomplish nearly the same thing. (Actually it does exactly what we want in this situation, so the implementation for Windows becomes nearly one line: ```  result = `where #{program.to_s}` ``` )  But the command shouldn't be called ```where``` either, for the same reason it shouldn't be called ```which```.  
The method name should reflect the behavior, not the platform-specific name (or implementation) for the behavior.  
So just as you've sketched out above, there could be platform-specific classes whose sole responsibility is to validate the command before it gets sent along to create the appropriate process.  I'd suggest that the name of the classes be ```[OS]CommandValidator``` or something like that.  

###### How does making a change like this fit into what you're doing now?   
 I don't think it should be made with this PR; I suspect that either you should do it as part of the re-working of Platform that you're doing now, or do this change after your work. I just want to try to keep work coordinated.  You and @mattwynne  certainly are the ones to make the call on that.

dg-ratiodata: I understand your concerns about the `which`. Let's move that discussion to a different thread.

lib/aruba/environments/unix_environment.rb
dg-ratiodata: Mmh... The order of the `<<` is important. `find` walks the collection "top down" and stops when the block returns true. I use this pattern in a few projects and it works. Not sure why this fails here. 

```ruby
objs = []
objs << 1
objs << 2

objs.find { |o| o > 0 }

objs = []
objs << 2
objs << 1

objs.find { |o| o > 0 }
```
Running this in pry:

```
[7] pry(main)> objs = []
=> []
[8] pry(main)> objs << 1
=> [1]
[9] pry(main)> objs << 2
=> [1, 2]
[10] pry(main)>
[11] pry(main)> objs.find { |o| o > 0 }
=> 1
[12] pry(main)>
[13] pry(main)> objs = []
=> []
[14] pry(main)> objs << 2
=> [2]
[15] pry(main)> objs << 1
=> [2, 1]
[16] pry(main)>
[17] pry(main)> objs.find { |o| o > 0 }
=> 2
```

weedySeaDragon: Makes sense.  Where I think my point of view is different from yours is that it seems to me that you're mixing *implementation details* into modeling your domain ( the real world: the environment variables).  Yes, this gives you the results you need, but now you're no longer actually reflecting what an Environment really is.  Doing it this way, you're saying: We always have a UnixEnvironment, and in some cases we have *both* a UnixEnvironment and a WindowsEnvironment -- because both can be true (both can 'match'). That's not correct.  It's always one or the other, but not both (not according to what aruba is going to know and interact with, anyway).  
From my point of view, it's always best to model how things really are.  If we need to use implementation details/artifacts to get things accomplished, then that should be cleanly and obviously separated.  (There are lots of ways to accomplish that.  Ports and adapters is one, as it happens.)
This is a case where it's easy to model things as they really are, so it should be done.

maxmeyer: I'm on gitter now.

mattwynne: My preference would just be to call `platform.environment_variables`. We only need to make the windows / unix decision once, and everything should fall out from that.

dg-ratiodata: Ok. Already done in #299.

dg-ratiodata: Ok, then merged #299 to have a foundation for @weedySeaDragon's work.

maxmeyer:
Cleanup. Closed for now.

PR #295, opened 2015-07-20
Try to fix problem with scopes

maxmeyer:
This is the first try to fix the problem with concurrent implementations of methods within `cucumber`. 

Known conflicts:
* all, within: `rspec`, `capybara` (jnicklas/capybara#1396, jnicklas/capybara#1547, jnicklas/capybara#1526)

Unfortunately this will break the existing API if someone wrote custom steps using the aruba-API. Now they also need to wrap their code within `aruba_scope {}`.

Outstanding: Real World Test with `middleman` (middleman/middleman#1562) if this really works.

This PR also includes some fixes, we need to merge separately if this PR does not solve the problem at all and is therefor dropped.

### DIFF
--- a/features/step_definitions/aruba_dev_steps.rb
+++ b/features/step_definitions/aruba_dev_steps.rb
@@ -26,7 +26,9 @@ Then /^the following step should fail with Spec::Expectations::ExpectationNotMet
 end
 
 Then /^the output should be (\d+) bytes long$/ do |length|
-  expect(all_output.length).to eq length.to_i
+  aruba_scope do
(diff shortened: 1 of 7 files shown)


mattwynne:
Maybe I'm wrong, because this isn't a problem I've experienced first-hand, but I think this breaks the 80/20 rule - most people most of the time should get the easy way of using Aruba. People who need to avoid clashing scopes should be able to opt *in* to this feature, but it should not be the default.

dg-ratiodata:
I decided to use just a aliased matcher for that. Hope this will prevent any further name clashes.

Issue #298, opened 2015-07-22
Make "Platform" to abstract OS-specific things


dg-ratiodata:
We have a few place where we need to do things differently. Let's introduce [this pattern](http://alistair.cockburn.us/Hexagonal+architecture) to `aruba` as suggested by @mattwynne in https://gitter.im/cucumber/aruba and discussed with @weedySeaDragon.

PR #299, opened 2015-07-22
Introducing the Ports & Adapters to `aruba`'s codebase

dg-ratiodata:
This is the PR for #298. It introduces three classes to aruba, which abstract the os specific code. @mattwynne Matt, is this what you've in mind? Pinging @weedySeaDragon as well.

I split this PR in several commits, each one can be reviewed by it's own.

### DIFF
--- a/lib/aruba/platform.rb
+++ b/lib/aruba/platform.rb
@@ -1,257 +1,29 @@
-require 'rbconfig'
-require 'pathname'
-
-require 'aruba/platform/simple_table'
+require 'thread'
+require 'aruba/platforms/unix_platform'
+require 'aruba/platforms/windows_platform'
 
 module Aruba
-  # WARNING:
-  # All methods found here are not considered part of the public API of aruba.
-  #
-  # Those methods can be changed at any time in the feature or removed without
-  # any further notice.
-  module Platform
-    def detect_ruby(cmd)
-      if cmd =~ /^ruby\s/
-        cmd.gsub(/^ruby\s/, "#{current_ruby} ")
-      else
-        cmd
-      end
+  PLATFORM_MUTEX = Mutex.new
+end
+
+module Aruba
+  class Platform < SimpleDelegator
+    def initialize
+      @platforms = []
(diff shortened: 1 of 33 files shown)


dg-ratiodata:
BTW: The jruby-failures seem to be somehow related to `0.8.0`... I hope to fix them soon.

lib/aruba/platform.rb
```diff
@@ -1,257 +1,29 @@
...
+      @platforms = []
+      @platforms << Aruba::Platforms::WindowsPlatform
+      @platforms << Aruba::Platforms::UnixPlatform
 
-    # Transform hash to a string table which can be output on stderr/stdout
-    def simple_table(hash)
-      SimpleTable.new(hash).to_s
+      super @platforms.find(&:match?).new
     end
+  end
```
mattwynne: The other way to do this is simply to do:

```
module Aruba
  Platform = [Platforms::WindowsPlatform, Platforms::UnixPlatform].find(&:match).new
end
```

lib/aruba/platform.rb
```diff
@@ -1,257 +1,29 @@
...
-    def write_file(path, content)
-      if RUBY_VERSION < '1.9.3'
-        File.open(path, 'wb') do |f|
-          f.print content
-        end
-      else
-        File.write(path, content)
-      end
-    end
+  PLATFORM_MUTEX = Mutex.new
```
mattwynne: What's this for?

mattwynne:
:claps:

lib/aruba/platform.rb
dg-ratiodata: Paranoia... I read something about class instance variables being not thread safe.

lib/aruba/platform.rb
dg-ratiodata: That's a pretty one. Fixed in master.

lib/aruba/platform.rb
mattwynne: That's only if you mutate things. We're keeping this as a constant now, so it should be fine. Let's just solve the problems we have in front of us! :)

Issue #291, opened 2015-07-16
Windows ENV variables are case-insensitive (but not to Ruby)


weedySeaDragon:
Windows treats its environment variables without regard to the CASE. That is, in Windows, the environment variable <code>'PATH' == 'Path' == 'PaTh' == 'patH'</code> etc.  Using <code>ENV</code> to query the environment variables from Windows is also case-insensitive.  So: <code>ENV['PATH'] == ENV['Path']</code>

However, once that information is brought into Ruby (e.g. by something like <code>environment = ENV.to_hash</code>, then case matters.  In this case, the keys in <code>environment</code> _are_ case sensitive.  So <code>environment['PATH'] </code> is _not_ equal to <code>environment['Path']</code>

This is causing failure anywhere that ENV['PATH'] is used.  (Specs and features that depend on this all fail under Windows.)  ENV['Path'] = _<the actual path information, according to Windows>_ is what is copied into <code>@environment</code>  But the key used is still 'Path' (mixed case).  But in the rest of the code, the key 'PATH' is used.  Since nothing exists for that key, things don't happen like they should.

I suggest that the use of ENV['PATH'] be isolated (turned in to a method call).  Then any initial case conversion can be done privately (e.g. change things to upper case before any comparison is done with ENV['PATH']).  
I'm not sure where such a method should go.  (I haven't explored it.)

weedySeaDragon:
I see that you just did a major refactoring, @dg-ratiodata , and that you created the Environment class.  This is another platform (OS) dependent item.  So perhaps your Environment class should do some simple delegation to either a WindowsEnvironment (which will handle the case-sensitivity stuff so that it's invisible to everyone else) or a UnixEnvironment (or whatever you want to call the one that is not Windows).

maxmeyer:
The class is a little bit older, actually. It was introduced with 0.7. or 08. And yes that's what I've in mind. :smile:

weedySeaDragon:
I've created a PR for this (#293). With that PR (and the one it depends on: #290 ), all but one of the specs passed.  That one that didn't (Aruba::Api files #expand_path when file_name contains fixtures "%" string..)  needed to have one change:  it was using <code>instance_double('Aruba::Environment')</code>.  But that wouldn't work.  Changing the mock from an <code>instance_double</code> to an <code>object_double</code> did the trick.  ...Except for ruby 1.9.2.  The configuration features are failing with that one. I'm not sure if or how this might have caused that.  
[...]

PR #300, opened 2015-07-25
Windows Environment Vars are converted to all upper case

weedySeaDragon:
addresses #291 
```WindowsEnvironmentVars < Environment```  converts all ENV keys to upper case on initialization.  

```WindowsPlatform``` now uses ```WindowsEnvironmentVars``` as its environment:
```ruby
    class WindowsPlatform < UnixPlatform
      ...
      def environment_variables
        WindowsEnvironmentVars.new
      end
```

Rspec tests for ```WindowsEnvironmentVars``` included to be sure that accessing the environment variables by the keys will work and that ```WindowsEnvironmentVars.keys``` only has upper case chars.

-  I put the class into aruba/lib/platforms  but perhaps it should go somewhere else
-  "```WindowsEnvironmentVars.keys``` only has upper case chars." really means that it only checks to be sure that none of the keys match ```/[a-z]+/```  I think that might not catch lower case unicode chars that don't get 'upper-cased' by ```[String].upcase```
  - Is this is worth including a gem that does unicode checking and 'upper-casing' so that we can be sure to check this? (like [unicode_utils](https://github.com/lang/unicode_utils))   Or should we wait until someone has an issue with it and then deal with it?

### DIFF
--- a/lib/aruba/platforms/windows_environment_vars.rb
+++ b/lib/aruba/platforms/windows_environment_vars.rb
@@ -1,3 +1,5 @@
+require 'aruba/environment'
+
 module Aruba
   #
   # Windows is case-insensitive when it accesses its environment variables.  That means that
@@ -28,7 +30,9 @@ module Aruba
   class WindowsEnvironmentVars < Environment
     def initialize
       super
-      @env = ENV.to_hash.dup.inject({}) { |new_env, (k,v)| new_env[k.to_s.upcase] = v; new_env }
+      # rubocop:disable Style/EachWithObject
+      @env = ENV.to_hash.dup.inject({}) { |new_env, (k,v)| new_env[k.to_s.upcase] = v; new_env } #compatible with ruby < 1.9
+      # rubocop:enable Style/EachWithObject
     end
   end
 end
--- a/lib/aruba/platforms/windows_environment_vars.rb
+++ b/lib/aruba/platforms/windows_environment_vars.rb
@@ -1,0 +1,34 @@
+module Aruba
+  #
+  # Windows is case-insensitive when it accesses its environment variables.  That means that
+  # at the Windows command line:
+  #  C:>set Path
+  #  C:>Path=.;.\bin;c:\rubys\ruby-2.1.6-p336\bin;
+  #  C:>set PATH
+  #  C:>Path=.;.\bin;c:\rubys\ruby-2.1.6-p336\bin;
+  #
+  # If you access environment variables through ENV, you can access values no matter the
+  # case of the key:
+  #  irb: > ENV["Path"]
+  #  => ".;.\bin;c:\rubys\ruby-2.1.6-p336\bin;"
+  #  irb: > ENV["Path"]
+  #  => ".;.\bin;c:\rubys\ruby-2.1.6-p336\bin;"
+  #
+  # But if you copy the ENV to a hash, Ruby treats the keys as case sensitive:
+  #  irb: > env_copy = ENV.to_hash
+  #  => {"ALLUSERSPROFILE"=>"C:\\ProgramData", "ANSICON"=>"119x1000 (119x58)", "ANSICON_DEF"=>"7",  APPDATA"=>"C:\\Users\\blorf\\AppData\\Roaming", ....}
+  #  irb: > env["Path"]
+  #  => ".;.\bin;c:\rubys\ruby-2.1.6-p336\bin;"
+  #  irb: > env["PATH"]
+  #  => nil
+  #
+  # Thus we turn all of the environment variable keys to upper case so that aruba is ensured that
+  # accessing environment variables with upper case keys will always work.
+  #
+  class WindowsEnvironmentVars < Environment
+    def initialize
+      super
+      @env = ENV.to_hash.dup.inject({}) { |new_env, (k,v)| new_env[k.to_s.upcase] = v; new_env }
+    end
+  end
+end
--- a/lib/aruba/platforms/windows_environment_vars.rb
+++ b/lib/aruba/platforms/windows_environment_vars.rb
@@ -27,6 +27,8 @@ module Aruba
   # Thus we turn all of the environment variable keys to upper case so that aruba is ensured that
   # accessing environment variables with upper case keys will always work.
   #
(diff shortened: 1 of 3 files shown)


lib/aruba/platforms/windows_environment_vars.rb
```diff
@@ -0,0 +1,40 @@
...
+  # Thus we turn all of the environment variable keys to upper case so that aruba is ensured that
+  # accessing environment variables with upper case keys will always work.
+  #
+  # TODO: investigate unicode characters that don't respond to [String].upcase
+  #
+  class WindowsEnvironmentVars < Environment
+    def initialize
+      super
+      # rubocop:disable Style/EachWithObject
+      @env = ENV.to_hash.dup.inject({}) { |new_env, (k,v)| new_env[k.to_s.upcase] = v; new_env } #compatible with ruby < 1.9
```
maxmeyer: Let's use `if RUBY_VERSION < ' 1.9' here. And make it obvious that we need to change this code if we drop support for ruby < 1.9.3. Please no `#dup` here. `inject` creates a new hash.

lib/aruba/platforms/windows_platform.rb
```diff
@@ -15,6 +16,10 @@ class WindowsPlatform < UnixPlatform
       def self.match?
         FFI::Platform.windows?
       end
+
+      def environment_variables
+        WindowsEnvironmentVars.new
```
maxmeyer: My personal preference: No abbreviations. So `Variables` not `Vars`.

spec/aruba/platform/windows_environment_vars_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'spec_helper'
+
+# create a value that does not exist in the collection. append a number to the original_str until not found
```
maxmeyer: Will think about this a bit.

lib/aruba/platforms/windows_environment_vars.rb
```diff
@@ -0,0 +1,40 @@
...
+  #  => ".;.\bin;c:\rubys\ruby-2.1.6-p336\bin;"
+  #  irb: > env["PATH"]
+  #  => nil
+  #
+  # Thus we turn all of the environment variable keys to upper case so that aruba is ensured that
+  # accessing environment variables with upper case keys will always work.
+  #
+  # TODO: investigate unicode characters that don't respond to [String].upcase
+  #
+  class WindowsEnvironmentVars < Environment
```
dg-ratiodata: Please rename `Environment` to `EnvironmentVariab les`.

lib/aruba/platforms/windows_environment_vars.rb
```diff
@@ -0,0 +1,40 @@
...
+    def initialize
```
dg-ratiodata:
~~~ruby
def initialize(env = ENV.to_hash)
  if RUBY_VERSION < '1.9.3'
    @env = env.inject({}) { |a, (k,v)| a[k.to_s.upcase] = v; a}
  else
    @env = env.each_with_object({}) { |(k,v), a| a[k.to_s.upcase] = v  }
  end
end
~~~

dg-ratiodata: Ah, and please change the `#initialize`-method for `UnixEnvironment` as well.

spec/aruba/platform/windows_environment_vars_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  not_found_value = str
+  i = 1
+  while coll.include?(not_found_value)
+    not_found_value = "#{not_found_value}-#{i}"
+    i += 1
+  end
+  not_found_value
+end
+
+describe WindowsEnvironmentVars do
```
dg-ratiodata: Let's make this a little bit simpler here.

spec/aruba/platform/windows_environment_vars_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  subject(:windows_env) { described_class.new }
+
+  let(:env_1)         { "EnvVarAdapt-1" }
+  let(:env_1_value)   { "EVA one one one" }
+  let(:env_2)         { "EnvVarAdapt-2" }
+  let(:env_2_value)   { "EVA two two two" }
+  let(:default_value) { "default value" }
+
+  it "doesn't have any lower-case letters" do
```
dg-ratiodata: Please use a different style for tests:

~~~ruby
describe '#[]' do
  context 'when environment contains uppercase variable' do
    let(:old_environment) { { 'MY_VARIABLE' => '1' } }

    context 'when uppercase key is given' do
      let(:variable) { 'MY_VARIABLE' }

      it { expect(windows_env[variable]).to eq '1' }
    end

    context 'when lowercase key is given' do
      let(:variable) { 'my_variable' }

      it { expect(windows_env[variable]).to eq '1' }
    end
  end
end
~~~

spec/aruba/platform/windows_environment_vars_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    expect(windows_env.to_h.select {|k, v| (k =~ /[a-z]+/) }).to be_empty
+  end
+
+  describe "only finds upper-case keys" do
+    before(:each) do
+      @orig_env = ENV.to_hash
```
dg-ratiodata: Let's not fiddle around with `ENV`. The logic in `Windows|UnixEnvironmentVariables` is independent of `ENV`.

spec/aruba/platform/windows_environment_vars_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      it "doesn't find key different mixed case" do
+        expect(windows_env[env_2]).to be_falsey
+      end
+
+      it "doesn't find what it shouldn't find" do
+        expect( windows_env[nonexistant(windows_env.to_h.keys)] ).to be_falsey
+      end
+    end
+
+    describe "key?" do
```
dg-ratiodata: Please re-use the tests for `#[]` with the same style here.

maxmeyer:
Closed in favor of #303

PR #303, opened 2015-08-08
Introduce Windows Environment

maxmeyer:
This uses code from #300 and introduces some more tests and cleanups. I would like to see if the tests pass.


maxmeyer:
Ashley @weedySeaDragon, I left you as author of this commit. Hope this is ok for you.

PR #304, opened 2015-08-09
Split up which for Unix and Windows

maxmeyer:
Unix and Windows need a separate logic to find executables in PATH. This splits up the original implementation and uses `Aruba.platform` to decide which implementation to use.

### DIFF
--- a/lib/aruba/platforms/unix_platform.rb
+++ b/lib/aruba/platforms/unix_platform.rb
@@ -3,6 +3,7 @@ require 'pathname'
 
 require 'aruba/platforms/simple_table'
 require 'aruba/platforms/unix_command_string'
+require 'aruba/platforms/unix_which'
 
 module Aruba
   # This abstracts OS-specific things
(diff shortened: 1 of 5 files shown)


maxmeyer:
We now only need to make sure, that the logic is correct on windows.

maxmeyer:
@weedySeaDragon I took your comments about the `#match?` into account. Now there's a default-class which just fails. All other classes use a matcher which checks specific things.

maxmeyer:
Ashley, would you mind to have a look into the windows implementation. It is still failing, but Appveyor-builds took too long to use them for fixing bugs.

weedySeaDragon:
I see 2 things immediately:
1.  The full path to ``cmd.exe`` is not necessary.  (Windows will always know where to find cmd.exe). So ``WindowsCommandString.initialize`` can be changed to:   `` __setobj__ format('cmd.exe /c "%s"',  cmd)``  
2.  You are only passing 1 argument to ChildProcess (the entire command string).  In Windows, [ChildProcess requires you to pass each of the string components as a separate argument:](https://github.com/jarib/childprocess/issues/85#issuecomment-93970168)  
```
process = ChildProcess.build("cmd.exe", "/c", "path\to\command\cat")
```
(also see the first comment in #283)

maxmeyer:
> 1. The full path to cmd.exe is not necessary. 

For now: Unfortunately yes... The current code is not wrapped with `with_environment`. This is part of another PR of mine... This will be removed if the PR is merged.

> 2. You are only passing 1 argument to ChildProcess (the entire command string). In Windows, 

Oh. Will fix this.

PR #305, opened 2015-08-11
Add console support to aruba

dg-ratiodata:
This adds a console to aruba which users can use as "sandbox". It uses `tmp/aruba` as temporary directory.

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -18,12 +18,18 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'ffi', '~> 1.9.10'
   s.add_runtime_dependency 'rspec-expectations', '>= 2.99'
   s.add_runtime_dependency 'contracts', '~> 0.9'
+  s.add_runtime_dependency 'thor', '~> 0.19'
 
   s.add_development_dependency 'bundler', '~> 1.10.2'
 
   s.rubygems_version = ">= 1.6.1"
-  s.required_ruby_version = '>= 1.8.7'
-  s.post_install_message = <<-EOS
+
+  if Aruba::VERSION >= '1'
+    s.required_ruby_version = '>= 1.9.3'
+  else
+    s.required_ruby_version = '>= 1.8.7'
+
+    s.post_install_message = <<-EOS
 Use on ruby 1.8.7
 * Make sure you add something like that to your `Gemfile`. Otherwise you will
   get cucumber > 2 and this will fail on ruby 1.8.7
@@ -32,6 +38,7 @@ Use on ruby 1.8.7
 
 With aruba >= 1.0 there will be breaking changes. Make sure to read https://github.com/cucumber/aruba/blob/master/History.md for 1.0.0
 EOS
+  end
 
   s.files            = `git ls-files`.split("\n")
   s.test_files       = `git ls-files -- {spec,features}/*`.split("\n")
--- a/bin/aruba
+++ b/bin/aruba
@@ -1,0 +1,7 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path('../../lib', __FILE__)
+
+require 'aruba/cli'
+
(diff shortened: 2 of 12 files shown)


dg-ratiodata:
I merged this although there are some build failures. I could reproduce them with jruby 1.7.19. They are gone if you use jruby 1.7.20. On Travis we are forced to use 1.7.19.

dg-ratiodata:
@mattwynne Now we've got an aruba console :smile: Unfortunately we are bitten by some jruby bug(s). See comment above.

mattwynne:
\o/

Can you exclude the jruby tests with a tag or similar?

dg-ratiodata:
Mmmh... Yes, I would like to do so, but is there something similar to `skip_this_scenario` in `cucumber` 1.3.x? `#pending` make the test suite fail as well.

dg-ratiodata:
`#skip_this_scenario` seems to be cucumber > 2 only.

dg-ratiodata:
And cucumber 2 requires ruby > 1.9.3. As we supported ruby 1.8.7, 1.9.2 by "chance" and made it official again using travis, we're stuck to `cucumber` 1.3.x at least for rubies < 1.9.3 till aruba >= 1.

dg-ratiodata:
BTW: I think, I'm going to prepare `0.9` to get some feedback on the structural changes I made. @aslakhellesoy This time with a proper announcement on cukes. Do you prefer to have a proper written announcement or is it ok to copy&paste the History.md-entry for the release? With the next iteration I would like to fix the Windows issues.

dg-ratiodata:
Ok. I added a hack to cucumber.yml to make it possible even for `cucumber` 1.3.x.

dg-ratiodata:
Ah... And what a about pre-releases. Should I announce them as well?

mattwynne:
@dg-ratiodata so many questions!

> is there something similar to skip_this_scenario in cucumber 1.3.x? #pending make the test suite fail as well.

I would use tags and profiles to exclude the scenarios from the jruby build. That's what we do in Cucumber.

> And cucumber 2 requires ruby > 1.9.3. As we supported ruby 1.8.7, 1.9.2 by "chance" and made it official again using travis, we're stuck to cucumber 1.3.x at least for rubies < 1.9.3 till aruba >= 1.

That sucks. I guess RSpec is the main driver for still supporting 1.8.7 yes? I think we maybe need to involve Myron in the conversation about dropping older Ruby version support. I'll ping him in https://github.com/cucumber/cucumber-ruby/issues/902

> Do you prefer to have a proper written announcement or is it ok to copy&paste the History.md-entry for the release?

History.md cut / paste is better than nothing, but a bit of narrative above it about the purpose of the release, and something about where you're headed next would be good for context.

I'd like you to write a blog post for http://cucumber.io/blog about Aruba and where you're headed with it. You can send blog posts as PRs adding files to [this directory](https://github.com/cucumber/website/tree/master/apps/dynamic/views/_posts).

I think we need to formalise when we write a blog post and when we just post on cukes. I'm sloppy about this at the moment.

dg-ratiodata:
> I would use tags and profiles to exclude the scenarios from the jruby build. That's what we do in Cucumber.

Yeah, found that out, too. :smile: https://github.com/cucumber/aruba/blob/master/cucumber.yml#L9

> I guess RSpec is the main driver for still supporting 1.8.7 yes?

Yes, but they will stick to aruba 0.6.x for now. He had quite a few problems upgrading `aruba` to the latest version. Some problems were related to external bugs/problems by libraries used by aruba, but some were "home made" by aruba. See https://github.com/rspec/rspec-core/pull/2028 for the discussion on this topic. It took too much of his free time. 

> I think we need to formalise when we write a blog post and when we just post on cukes. I'm sloppy about this at the moment.

That would be great. Plus some guidelines where those post should be published... I've got a blog on my own ;-) I use regularly - see https://www.fedux.org/articles.html

PR #311, opened 2015-10-17
Try to make jruby 9000 work

maxmeyer:
This PR should make aruba compatible with jruby 9000.

Moved code here to prevent cluttering IRC and cukes@....


dg-ratiodata:
Seems to be fix by jruby itself. See succeeding PR #385.

PR #314, opened 2015-10-30
Fix '"#exit_timeout" is deprecated' error

yasuhito:
This PR replaces `exit_timeout` with `aruba.config.exit_timeout`.


dg-ratiodata:
Thanks a lot.

PR #316, opened 2015-11-18
Set signal used to stop/terminate a command

dg-ratiodata:
I need to run a docker container with `aruba`. This container is using `systemd` as PID 1. To shutdown the container after a test correctly I need to send it SIGINT instead of SIGTERM. This PR adds a configuration option to send an arbitrary signal to a command started. This is used in `command.terminate` which itself is used to cleanup processes after a test run. This PR also adds announcers.


mattwynne:
It's great that you've added docs for this. How come there are both scenarios about stopping the comment in the `run` feature, and a whole other set in the `stop` feature?

maxmeyer:
@mattwynne Sorry, I should have added WIP. I use the PRs to check against ruby < latest and jruby since I'm running Arch Linux with only the latest stable ruby available, but I don't like rvm etc.

The feature files you see are the documentation for ...

1. Aruba Ruby API: `features/api/command/*.feature`
2. Cucumber Steps: `features/steps/command/*.feature`

As you might have seen, I use `rspec` within the tests to test the API and I use `cucumber` to test aruba's cucumber steps. There are also quite a few changes in my local repository unpushed. I still need to move the tests for 1. from `features/api/command/run.feature` to `features/api/command/stop.feature` though I need to decide how to name the feature tests. 

Regarding the documentation: I'm open to suggestions. There are API methods to access the last command started and the like. Via those accessors, you get back the command started last. A command can be started, stopped, send a signal etc. I was not keen to add more and more methods to aruba to make things possible. Maybe we could discuss this in March and then prepare a 1.0.0 release with a major cleanup.

~~~ruby
last_command_started.send_signal
last_command_started.terminate
## [...]
~~~

I still face some severe issues with jruby ... It's always the same with this kind of ruby runtime. :-( Issues are problems with flushing IO and forking etc. most times... Now: It doesn't want to write text to stdout on receiving a signal which works fine with MRI... But I hope to get that stuff fixed soon - without eating my keyboard...

dg-ratiodata:
@mattwynne Sorry, I should have added WIP. I use the PRs to check against ruby < latest and jruby since I'm running Arch Linux with only the latest stable ruby available, but I don't like rvm etc.

The feature files you see are the documentation for ...

1. Aruba Ruby API: `features/api/command/*.feature`
2. Cucumber Steps: `features/steps/command/*.feature`

As you might have seen, I use `rspec` within the tests to test the API and I use `cucumber` to test aruba's cucumber steps. There are also quite a few changes in my local repository unpushed. I still need to move the tests for 1. from `features/api/command/run.feature` to `features/api/command/stop.feature` though I need to decide how to name the feature tests. 

Regarding the documentation: I'm open to suggestions. There are API methods to access the last command started and the like. Via those accessors, you get back the command started last. A command can be started, stopped, send a signal etc. I was not keen to add more and more methods to aruba to make things possible. Maybe we could discuss this in March and then prepare a 1.0.0 release with a major cleanup.

~~~ruby
last_command_started.send_signal
last_command_started.terminate
## [...]
~~~

I still face some severe issues with jruby ... It's always the same with this kind of ruby runtime. :-( Issues are problems with flushing IO and forking etc. most times... Now: It doesn't want to write text to stdout on receiving a signal which works fine with MRI... But I hope to get that stuff fixed soon - without eating my keyboard...

dg-ratiodata:
I'm not that happy with the documentation and I think I will need to refactor the structure to make it easier to read/use. But for now I'm adding documentation to the given structure.

maxmeyer:
@mattwynne JFYI: The problems I face seem to be related to `bundler`. Running a ruby script with the new background steps takes "too" long because `ruby` loads the `Gemfile` from the `aruba` project root. So if I try to send a signal to the command the receiver is not the script, but `bundler` or even `ruby` itself. 

Waiting at least 0.5s after starting a command fixes the problem "locally". Therefor: I added a "start up time"-configuration option to wait a few miliseconds/seconds after starting a command (in background) - defaults to "0".

Issue #69, opened 2011-04-16
Add run_script_string(:engine=> 'ruby', :cmd => string)


hedgehog:
Currently, you have to create a file, say `script_file`, then run `chmod +x script_file`, then call `run_simple("./script_file")`
Instead it would be useful to simply do:

```
cmd =<<-END_CMD.gsub(/^ {4}/, '')
  #!/usr/bin/env bash
  rvm gemset use vagrant
  cd #{vagrant.path}
  echo vagrant up #{vm_name}
  END_CMD
run_script_string(cmd)
stdout_from(cmd).should_not include(name)
```

of course the hash-bang could ruby etc.

In case it helps work out where any changes are required, the error you currently get:

```
    /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/childprocess-0.1.8/lib/childprocess/unix/process.rb:92:in `exec': No such file or directory - #!/usr/bin/env (Errno::ENOENT)
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/childprocess-0.1.8/lib/childprocess/unix/process.rb:92:in `block in launch_process'
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/childprocess-0.1.8/lib/childprocess/unix/process.rb:83:in `fork'
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/childprocess-0.1.8/lib/childprocess/unix/process.rb:83:in `launch_process'
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/childprocess-0.1.8/lib/childprocess/abstract_process.rb:52:in `start'
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/aruba-0.3.6/lib/aruba/process.rb:22:in `run!'
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/aruba-0.3.6/lib/aruba/api.rb:212:in `block in run'
from /home/hedge/.rvm/gems/ruby-1.9.2-p180@bbb_vm_monitor/gems/aruba-0.3.6/lib/aruba/api.rb:9:in `chdir'
[...]

PR #323, opened 2015-12-07
Features/run shell command

AdrieanKhisbe:
Add two new steps to run script or shell commands:
- `When I run the following script:``
- `When(/^I run the following (bash|zsh|fish|dash)? ?commands?:$/)`

cf  https://github.com/cucumber/aruba/issues/69#issuecomment-162305537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,5 +39,11 @@ notifications:
 env:
   global:
    - secure: l8uznA5K4K9mZ1krmP3lTMD8WcJ32qGxFOR3jubKHcOBSLB4xSzU2aIqjyJdO+rLzebkwamhJc8pGSIWOUDQYvFiX7splK+uEkbBJ5huAhXtLF4Qgl86bCWbEXYzN7rvn0DQfpJAovyFMNRMnfo70XhwqWzFsaYa7Z0YbqYsJE4=
-   # - JRUBY_OPTS='-Xcli.debug=true --debug'
-   - JRUBY_OPTS='--dev'
+   - JRUBY_OPTS='-Xcli.debug=true --debug'
+cache: apt
+addons:
+  apt:
+    packages:
+    - dash
+    - fish
+    - zsh
--- a/features/steps/command/shell.feature
+++ b/features/steps/command/shell.feature
@@ -62,7 +62,7 @@ Feature: Running shell commands
       Scenario: Running bash commands
         When I run the following commands with `bash`:
         \"\"\"bash
-        echo "Hello \c"
+        echo -n "Hello "
         echo `echo bash` # subshell
         \"\"\"
         Then the output should contain exactly "Hello bash"
@@ -98,7 +98,7 @@ Feature: Running shell commands
     """
     Feature: Running fish scripts
       Scenario: Running fish commands
-        When I run the following commands with `fish`:
+        When I run the following commands in `fish`:
         \"\"\"bash
         echo -n "Hello "
         echo (echo fish)
--- a/features/steps/command/shell.feature
+++ b/features/steps/command/shell.feature
@@ -1,0 +1,117 @@
+Feature: Running shell commands
+
+  You can run an *ad hoc* script with the following step
+  - `When I run the following script:`
+
+  Or you can run shell commands with:
+  - `I run the following (bash|zsh|fish|dash)? commands`
+
+  Background:
+    Given I use a fixture named "cli-app"
+
+  Scenario: Running ruby script
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running scripts
+      Scenario: Running ruby script
+        When I run the following script:
+        \"\"\"bash
+        #!/usr/bin/env ruby
+
+        puts "Hello"
+        \"\"\"
+        Then the output should contain exactly "Hello"
+    """
+    When I run `cucumber`
+    Then the features should all pass
+
+  Scenario: Running python script
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running scripts
+      Scenario: Running python script
+        When I run the following script:
+        \"\"\"bash
+        #!/usr/bin/env python
+
+        print("Hello")
+        \"\"\"
+        Then the output should contain exactly "Hello"
+    """
+    When I run `cucumber`
+    Then the features should all pass
+
+  Scenario: Running shell commands
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running scripts
+      Scenario: Running shell commands
+        When I run the following commands:
+        \"\"\"bash
+        echo "Hello shell"
+        \"\"\"
+        Then the output should contain exactly "Hello shell"
+    """
+    When I run `cucumber`
+    Then the features should all pass
+
+  Scenario: Running bash commands
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running scripts
+      Scenario: Running bash commands
+        When I run the following commands with `bash`:
+        \"\"\"bash
+        echo "Hello \c"
+        echo `echo bash` # subshell
+        \"\"\"
+        Then the output should contain exactly "Hello bash"
+    """
+    When I run `cucumber`
+    Then the features should all pass
+
+  Scenario: Running zsh commands
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running zsh scripts
+      Scenario: Running zsh commands
+        When I run the following commands in `zsh`:
+        \"\"\"bash
+        echo "Hello \c"
+        echo $((2 + 2))
+        \"\"\"
+        Then the output should contain exactly "Hello 4"
+
+      Scenario: Running zsh commands
+        When I run the following commands with `/bin/zsh`:
+        \"\"\"bash
+        echo "Hello \c"
+        echo $((6 - 2))
+        \"\"\"
+        Then the output should contain exactly "Hello 4"
+    """
+    When I run `cucumber`
+    Then the features should all pass
+
+  Scenario: Running fish commands
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running fish scripts
+      Scenario: Running fish commands
+        When I run the following commands with `fish`:
+        \"\"\"bash
+        echo -n "Hello "
+        echo (echo fish)
+        \"\"\"
+        Then the output should contain exactly "Hello fish"
+
+      Scenario: Running fish commands with explicit path
+        When I run the following commands with `/usr/bin/env fish`:
+        \"\"\"bash
+        echo -n "Hello "
+        echo (echo fish)
+        \"\"\"
+        Then the output should contain exactly "Hello fish"
+    """
+    When I run `cucumber`
+    Then the features should all pass
--- a/lib/aruba/cucumber/command.rb
+++ b/lib/aruba/cucumber/command.rb
@@ -1,6 +1,7 @@
 if Aruba::VERSION < '1.0.0'
   require 'aruba/cucumber/core'
 end
+require 'aruba/generators/script_file'
 
 When(/^I run "(.*)"$/)do |cmd|
   warn(%{\e[35m    The /^I run "(.*)"$/ step definition is deprecated. Please use the `backticks` version\e[0m})
@@ -26,6 +27,17 @@ end
 When(/^I successfully run `(.*?)`(?: for up to (\d+) seconds)?$/)do |cmd, secs|
   cmd = sanitize_text(cmd)
   run_simple(cmd, :fail_on_error => true, :exit_timeout => secs && secs.to_i)
+end
+
+When(/^I run the following (?:commands|script)(?: (?:with|in) `([^`]+)`)?:$/) do |shell, commands|
+  prepend_environment_variable('PATH', expand_path('bin') + ':')
+
+  Aruba.platform.mkdir(expand_path('bin'))
+  shell ||= Aruba.platform.default_shell
+
+  Aruba::ScriptFile.new(:interpreter => shell, :content => commands,
+                 :path => expand_path('bin/myscript')).call
+  step 'I run `myscript`'
 end
 
 When(/^I run "([^"]*)" interactively$/) do |cmd|
--- a/lib/aruba/generators/script_file.rb
+++ b/lib/aruba/generators/script_file.rb
@@ -1,0 +1,41 @@
+# Aruba
+module Aruba
+  # ScriptFile
+  class ScriptFile
+    def initialize(opts = {})
+      @path        = opts[:path]
+      @content     = opts[:content]
+      @interpreter = opts[:interpreter]
+    end
+
+    def call
+      Aruba.platform.write_file(@path, "#{header}#{@content}")
+      Aruba.platform.chmod(0755, @path, {})
+    end
+
+    private
+
+    def header
+      if script_starts_with_shebang?
+        ''
+      elsif interpreter_is_absolute_path?
+        format("#!%s\n", @interpreter)
+      elsif interpreter_is_just_the_name_of_shell?
+        format("#!/usr/bin/env %s\n", @interpreter)
+      end
+    end
+
+    def interpreter_is_absolute_path?
+      Aruba.platform.absolute_path? @interpreter
+    end
+
+    def interpreter_is_just_the_name_of_shell?
+      @interpreter =~ /^[-_a-zA-Z.]+$/
+    end
+
+    def script_starts_with_shebang?
+      @content.start_with? '#!'
+    end
+
+  end
+end
--- a/lib/aruba/platforms/unix_platform.rb
+++ b/lib/aruba/platforms/unix_platform.rb
@@ -78,6 +78,10 @@ module Aruba
         LocalEnvironment.new.call(env, &block)
       end
 
+      def default_shell
+        'bash'
+      end
+
       def detect_ruby(cmd)
         if cmd =~ /^ruby\s/
           cmd.gsub(/^ruby\s/, "#{current_ruby} ")


SteveDonie:
Please think about running on Windows also.

AdrieanKhisbe:
@SteveDonie, would be usefull indeed, but it might need a dedicated step for the *command*,
and I'm not familiar with ruby&windows integration. (cmd, powershell)

(note that the windows test were already broken before that, and my tests don't seems to be even run)

dg-ratiodata:
@AdrieanKhisbe Windows Compatibility is on the TODO list but not with a high priority for me. If anyone likes to work on that, I'm happy to review/merge PRs.

dg-ratiodata:
:+1+ For the feature from my side as a whole. Now comming to the review.

dg-ratiodata:
Please ...

1. ... fix the comments I added
2. ... merge your PR into 2 commits: One for the tests + setup of test infrastructure and one for the implementation

features/steps/command/shell.feature
```diff
@@ -0,0 +1,76 @@
+Feature: Running shell commands
+
+  In order to test interactive command line applications
```
dg-ratiodata: I used that style myself in some features, but move away from this:

We want to generate documentation from the feature files in the near future. So please write some more "prosa" in natural language you would normally expect in documentation.

See https://github.com/cucumber/aruba/blob/master/features/api/command/run.feature for an example. Hope it's clear what I mean.

features/steps/command/shell.feature
```diff
@@ -0,0 +1,76 @@
...
+    Given I use a fixture named "cli-app"
+
+  Scenario: Running script
+    Given a file named "features/sehll.feature" with:
+
+    """
+   Feature: Running scripts
+    Scenario: Running ruby script
+      When I run the following script:
+        \"\"\"bash
```
dg-ratiodata: Please make sure, that indentation is similar to https://github.com/cucumber/aruba/blob/master/features/api/command/run.feature.

features/steps/command/shell.feature
```diff
@@ -0,0 +1,76 @@
...
+        #!/usr/bin/env ruby
+
+      puts "Hello"
+      \"\"\"
+      Then the output should contain exactly "Hello"
+
+
+     Scenario: Running python script
+     When I run the following script:
+        \"\"\"bash
```
dg-ratiodata: Same here.

lib/aruba/cucumber/command.rb
```diff
@@ -28,6 +28,21 @@
   run_simple(cmd, :fail_on_error => true, :exit_timeout => secs && secs.to_i)
 end
 
+When(/I run the following script:$/) do |file_content|
+  prepend_environment_variable('PATH', expand_path('bin') + ':')
```
dg-ratiodata: Is `#prepend_environment_variable` really needed here? It's added by default here: https://github.com/cucumber/aruba/blob/master/lib/aruba/cucumber/hooks.rb#L17.

lib/aruba/cucumber/command.rb
```diff
@@ -28,6 +28,21 @@
...
+  step 'an executable named "./bin/myscript" with:', file_content
+  step 'I run `myscript`'
+end
+
+When(/^I run the following (bash|zsh|fish|dash)? ?commands?:$/) do |shell_type, file_content|
```
dg-ratiodata: 1. Style of regex
  We used the `(?:(bash|zsh|fish|dash) )?` such a thing in other steps as well. Would be great if you make yours look the same. Something similar to this:
  
  ~~~
  When(/^I run the following (?:(bash|zsh|fish|dash) )?commands?:$/)
  ~~~

2. Explicit state the shell
  
  Is it really required (for your use case) to tell what shell is used or can we use `ENV['SHELL']` instead?

  @SteveDonie Do you know if this works on windows too?

3. Extract "generator" for scripts into class(es)

  Given we need to support windows too, I think it might make sense to extract the logic in to a `ScriptGenerator`-classes, maybe even a single class for each shell type or just a single one for windows/unix.
  
  * UnixScriptGenerator
  * WindowsScriptGenerator < UnixScriptGenerator

features/steps/command/shell.feature
```diff
@@ -0,0 +1,76 @@
...
+    Then the features should all pass
+
+  Scenario: Running shell commands
+    Given a file named "features/shell.feature" with:
+     """
+     Feature: Running shell commands
+
+       Scenario: Running shell commands
+        When I run the following commands:
+        \"\"\"bash
```
dg-ratiodata: Please fix as well.

features/steps/command/shell.feature
```diff
@@ -0,0 +1,76 @@
...
+           echo "Hello shell"
+        \"\"\"
+        Then the output should contain exactly "Hello shell"
+
+      Scenario: Running bash commands
+        When I run the following bash commands:
+          \"\"\"bash
```
dg-ratiodata: Please fix as well.

features/steps/command/shell.feature
```diff
@@ -0,0 +1,76 @@
...
+         echo -n "Hello "
+         echo `echo bash` # subshell
+          \"\"\"
+        Then the output should contain exactly "Hello bash"
+
+      Scenario: Running zsh commands
+        When I run the following zsh commands:
+          \"\"\"bash
```
dg-ratiodata: Please fix as well.

AdrieanKhisbe:
Thanks for your feedback @dg-ratiodata !! :)

Will apply your recommandation later today or tomorrow. (being apidays....)

lib/aruba/cucumber/command.rb
AdrieanKhisbe: Yep, without the executable is not found.

lib/aruba/cucumber/command.rb
AdrieanKhisbe: About point 2: 
Yep, I need to be explicity saying what is the shell I'm running command into to test compatibility of commands, cli-under-test with the different shell dialects.
(I'm using aruba to set up acceptance testing for  [resty](https://github.com/micha/resty))

AdrieanKhisbe: About point 3: Where should I put the generator class?

AdrieanKhisbe:
@dg-ratiodata  Applied your recommandations

Pending points: 
- Point 2: 
    Yep, I need to be explicity saying what is the shell I'm running command into to test compatibility of     commands, cli-under-test with the different shell dialects.
     (I'm using aruba to set up acceptance testing for resty)
- point 3: Where should I put the generator class you suggest, and what should be it's interface?

lib/aruba/cucumber/command.rb
dg-ratiodata: Mmmh... Can you try to invoke your script with `./bin/myscript` in the run step and leave out the prepend api call? If I remember correctly aruba accepts relative paths.

dg-ratiodata:
Let's try `lib/aruba/generators/script_file.rb` with something like this. Have a look at `lib/arub/command.rb` for another example of this pattern. I just wrote it down, that's not well formatted code. :

~~~ruby
# [...]
class ScriptFile
  class ZshFile
    def self.match?(matcher)
    end
  end
  
  class NullShellFile
     def self.match?(*)
        raise ArgumentError, 'Invalid input'
    end
  end
  def initialize(opts = {})
    @generators = []
    @generators << ZshFile
    @generators << ...
    @generators << NullShellFile

    @generator = @generators.find { |g| g.match? opts[:generator] }
  end
  def call
    @generator.call
  end
end
~~~

AdrieanKhisbe:
@dg-ratiodata can you tell me how the generator will be used.
I'll have a better Idea what I should put in the body of the `self.match?` :)

dg-ratiodata:
I would place `ShellFile.new(path).call` here https://github.com/cucumber/aruba/pull/323/files#diff-c3286c2e7f8674e13bb73a8f97895387R40.

lib/aruba/cucumber/command.rb
AdrieanKhisbe: Nope :/, 
I get a `Command "./bin/myscript" not found in PATH-variable`

AdrieanKhisbe:
@dg-ratiodata 
Some confirmation requests/questions so that I see clear:
- each generator willl have a `call` method, that will be responsible to create the file?
- this method should call `def write_file(name, content)` from `file_system.rb`
- how do I retrieve the content that should be put in the file?

AdrieanKhisbe:
> $ touch @dg-ratiodata # :smiley:

dg-ratiodata:
Sorry @AdrieanKhisbe I was on vacation.

> each generator willl have a call method, that will be responsible to create the file?

Yes.

> this method should call def write_file(name, content) from file_system.rb

No. This generator is internally used by the API and does not use its public methods. If you need to expand the path expand in via `expand_path` in the cucumber step and pass the expanded path into the generator. That's what I would do.

> how do I retrieve the content that should be put in the file?

I would use an instance variable for the specific generators which hold the "template". I think for the time being this should be just expanding a string. If we need to expand the support we could either think about using Erubis or even `Thor` as file generator.

AdrieanKhisbe:
@dg-ratiodata I worked a draft following what you tell me. you can see it in 25466d0.

I didn't get it working yet. (for some reason the step are not detected:

```
    +You can implement step definitions for undefined steps with these snippets:
      +
      +When(/^I run the following commands:$/) do |string|
      +  pending # Write code here that turns the phrase above into concrete actions

```

Thanks in advance for your feedback :)

lib/aruba/cucumber/command.rb
maxmeyer: Mhh... That's a thing we should have a look into. :-/

.travis.yml
```diff
@@ -39,3 +39,10 @@ env:
   global:
    - secure: l8uznA5K4K9mZ1krmP3lTMD8WcJ32qGxFOR3jubKHcOBSLB4xSzU2aIqjyJdO+rLzebkwamhJc8pGSIWOUDQYvFiX7splK+uEkbBJ5huAhXtLF4Qgl86bCWbEXYzN7rvn0DQfpJAovyFMNRMnfo70XhwqWzFsaYa7Z0YbqYsJE4=
    - JRUBY_OPTS='-Xcli.debug=true --debug'
+cache: apt
+addons:
+  apt:
+    packages:
+    - dash
+    - fish
+    - zsh
```
maxmeyer: The newline is missing at the end of file. Normally your editor should add this by default. Please fix this

features/steps/command/shell.feature
```diff
@@ -0,0 +1,73 @@
...
+      Scenario: Running fish commands
+        When I run the following fish commands:
+        \"\"\"bash
+        echo -n "Hello "
+        echo (echo fish)
+        \"\"\"
+        Then the output should contain exactly "Hello fish"
+    """
+    When I run `cucumber`
+    Then the features should all pass
```
maxmeyer: Same here. Newline is missing.

AdrieanKhisbe:
Thanks for your recommandations @maxmeyer,
Applied them in 4f0ef22

maxmeyer:
@AdrieanKhisbe @maxmeyer (private) and @dg-ratiodata (corporate) are the same person, me ;-) So don't be confused.

AdrieanKhisbe:
haha. I had some doubt according to logo.
But was too early for me to have a :bulb: eureka ^^

AdrieanKhisbe:
> $ touch this # :)

features/steps/command/shell.feature
```diff
@@ -0,0 +1,73 @@
+Feature: Running shell commands
+
+  You can run an *ad hoc* script with the following step
+  - `When I run the following script:`
+
+  Or you can run shell commands with:
+  - `I run the following (bash|zsh|fish|dash)? commands
```
dg-ratiodata: Backtick (End) is missing here.

AdrieanKhisbe: noted

AdrieanKhisbe: (tell me if it's the only thing to be fixed, I'll commit and push right away :))

lib/aruba/generators/script_file.rb
```diff
@@ -0,0 +1,79 @@
+# Aruba
+module Aruba
+  # ScriptFile
+  class ScriptFile
+
```
dg-ratiodata: please remove empty lines. Otherwise `rubocop` will complain and the build will fail. ;-)

AdrieanKhisbe:
Indeed, `rubocop` was not hapi :)

lib/aruba/generators/script_file.rb
```diff
@@ -0,0 +1,79 @@
...
+      def call
+        text = "#{header}\n#{@content}"
+        Aruba.platform.write_file(@path, text)
+        Aruba.platform.chmod('0755', @path)
+      end
+
+    end
+
+    class ZshFile < AbstractScript
+      def self.name
```
dg-ratiodata: Don't forget to make name an instance method as well.

features/steps/command/shell.feature
```diff
@@ -0,0 +1,73 @@
...
+
+  Background:
+    Given I use a fixture named "cli-app"
+
+  Scenario: Running script
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running scripts
+      Scenario: Running ruby script
+        When I run the following script:
```
dg-ratiodata: I know it's a little bit annoying for you :sweat_smile:, but what do you think
about the following idea: Let's merge the new steps you want to add into a
single one.

~~~ruby
When(/^I run the following commands(?: with `([^`]+))?`:$/) do |shell, commands|
  prepend_environment_variable('PATH', expand_path('bin') + ':')

  shell ||= Aruba.platforms.default_shell # => should return 'bash' on unix

  # Be care full I change generator to interpreter here
  ScriptFile.new(:interpreter => shell, :content => file_content,
                 :path => expand_path('bin/myscript')).call
  step 'I run `myscript`'
end
~~~

If we support partial strings, users can pass `/usr/bin/env ruby` to the step
and the correct generator is found for this too. If the user passes
`/usr/bin/bash` to the step, this would be used instead of `/usr/bin/env bash`. Would
this work for your use case? If you like that idea and implement it, can you
also add cucumber examples for that?

~~~ruby
class AbstractScript
  private

  attr_reader :path, :content, :interpreter

  public

  def initialize(path, content, interpreter)
    @path        = path
    @content     = content
    @interpreter = interpreter
  end

  def self.match?(matcher)
    matcher =~ Regexp.new(shell)
  end

  def call
    h = if interpreter_is_absolute_path
          format('#! %s', interpreter)
        elsif interpreter_is_just_the_name_of_shell
          header
        else
          format('#! %s', interpreter)
        end

    text = "#{h}\n#{content}"

    Aruba.platform.write_file(path, text)
    Aruba.platform.chmod('0755', path)
  end

  private

  def interpreter_is_absolute_path
    Aruba.platform.absolute_path? interpreter
  end

  def interpreter_is_just_the_name_of_shell?
    interpreter == shell
  end
end
~~~


What do you think about that? Using this code we could get rid of the
`self.name` method. Personally I like that idea. ;-)

~~~ruby
class AbstractScript
  def self.shell
    name.gsub(/^([^:]+::)/, '').gsub(/File$/, '').downcase
  end

  def shell
    self.shell
  end
end
~~~

dg-ratiodata: The code above is untested and might need to be fixed.

dg-ratiodata: And please rebase and fix the merge conflicts.

AdrieanKhisbe: Seems like a good idea to get ride of the `name` this way!!

About merging the two step i'm not sure. Having just to copy the script there without modifying 
the shebang would be better, and more clear on read

(what code section are you targeting?)

dg-ratiodata:
@SteveDonie It looks like we will have a good foundation in this PR for supporting windows as well. If the work here has been finished here, would you be willing to add Windows support - cmd.exe and power shell to `aruba`? Are you an active Windows user?

TODOs
* Add generators for cmd.exe and powershell
* Add `#default_shell` for windows to `lib/aruba/platforms/windows_platform.rb`
* ...

features/steps/command/shell.feature
dg-ratiodata: I'm going to send you a PR the next couple of hours to show you what I'm mean.

AdrieanKhisbe: Okey dokey :)

AdrieanKhisbe:
@dg-ratiodata, I updated the test (thought i don't manage to make them work so far)

AdrieanKhisbe:
> undefined method `platforms' for Aruba:Module (NoMethodError)

and also ScriptFile constant not defined -> where should I include the generator?

dg-ratiodata:
@AdrieanKhisbe Added some more comments to clarify what I mean. Thanks for your help, BTW. :-)

AdrieanKhisbe:
Applied most of them, but still don't know where I  should include the generator :)

lib/aruba/cucumber/command.rb
```diff
@@ -28,6 +29,17 @@
   run_simple(cmd, :fail_on_error => true, :exit_timeout => secs && secs.to_i)
 end
 
+When(/^I run the following commands(?: (?:with|in) `([^`]+)`)?:$/) do |shell, commands|
```
dg-ratiodata: Old
~~~
When(/^I run the following commands(?: (?:with|in) `([^`]+)`)?:$/) 
~~~

Suggestion for new
~~~
When(/^I run the following (?:commands|script)(?: (?:with|in) `([^`]+)`)?:$/)
~~~

AdrieanKhisbe: :+1:

lib/aruba/generators/script_file.rb
```diff
@@ -0,0 +1,37 @@
...
+  # ScriptFile
+  class ScriptFile
+    def initialize(opts = {})
+      @path        = opts[:path]
+      @content     = opts[:content]
+      @interpreter = opts[:interpreter]
+    end
+
+    def call
+      script_content = @content.start_with? '#!' ? @content : "#{header}\n#{@content}"
```
dg-ratiodata: Please move to `def header`:

~~~ruby
if script_starts_with_shebang?
  ''
elsif interpreter_is_absolute_path?
# [...]
~~~

~~~ruby
private

# [...]
def script_starts_with_shebang?
  @content.start_with? '#!' 
end
~~~

And also move the "\n" to the `#format`-methods. Then we can just `#{header}#{content}`.
~~

features/steps/command/shell.feature
```diff
@@ -0,0 +1,66 @@
...
+
+        puts "Hello"
+        \"\"\"
+        Then the output should contain exactly "Hello"
+
+      Scenario: Running python script
+        When I run the following commands:
+        \"\"\"bash
+        #!/usr/bin/env python
+
```
dg-ratiodata: Splitting up those tests on the "highest" level makes it easier to troubleshooting our test suite.
~~~ cucumber
Scenario: Running ruby script
  Given a file named "features/shell.feature" with:
  """
  Feature: Running script
    Scenario: Running script
      When I run the following commands:
      \"\"\"bash
      #!/usr/bin/env ruby

      puts "Hello"
      \"\"\"
      Then the output should contain exactly "Hello"
~~~

~~~ cucumber
Scenario: Running python script
  Given a file named "features/shell.feature" with:
  """
  Feature: Running script
    Scenario: Running script
      When I run the following commands:
      \"\"\"bash
      #!/usr/bin/env python

      print("Hello")
      \"\"\"
      Then the output should contain exactly "Hello"
~~~

AdrieanKhisbe:
@dg-ratiodata  \o/
The test is passsiiiing ;D

lib/aruba/generators/script_file.rb
```diff
@@ -0,0 +1,44 @@
...
+      Aruba.platform.write_file(@path, script_content)
+      Aruba.platform.chmod(0755, @path, {})
+    end
+
+    private
+
+    def header
+      if script_starts_with_shebang?
+        ''
+      else
```
maxmeyer: elsif + no nesting please.

AdrieanKhisbe: okeydokey

features/steps/command/shell.feature
```diff
@@ -0,0 +1,101 @@
...
+    """
+    When I run `cucumber`
+    Then the features should all pass
+
+  Scenario: Running zsh commands
+    Given a file named "features/shell.feature" with:
+    """
+    Feature: Running zsh scripts
+      Scenario: Running zsh commands
+        When I run the following commands in `zsh`:
```
maxmeyer: It's ok to add another inner scenario for `with` and `in` just to make sure that users directly see the difference in wording.

lib/aruba/generators/script_file.rb
AdrieanKhisbe: (don't know why that occured ^^)

features/steps/command/shell.feature
AdrieanKhisbe: Okey dokey :)

maxmeyer:
Besides the comments. Great mate. We're almost done! :-) After you fixed the things mentioned in the comments, please squash everything in to two commits. One for the tests and one for the implementation.

maxmeyer:
It's best to fix, then push and then rebase to squash the commits. But maybe you already know that. If something went wrong you've got a backup online.

features/steps/command/shell.feature
AdrieanKhisbe: voila

AdrieanKhisbe:
Tell me if it's okey and I rebase/squash

lib/aruba/generators/script_file.rb
```diff
@@ -0,0 +1,42 @@
...
+  # ScriptFile
+  class ScriptFile
+    def initialize(opts = {})
+      @path        = opts[:path]
+      @content     = opts[:content]
+      @interpreter = opts[:interpreter]
+    end
+
+    def call
+      script_content = script_starts_with_shebang? ? @content : "#{header}#{@content}"
```
maxmeyer: :-) No need for this test. Please remove line 12 and change the code like this.

~~~
Aruba.platform.write_file(@path, "#{header}#{@content}")
~~~

maxmeyer:
One last thing and you can rebase. See my last comment. You duplicated a test which is not neccessary. The same check is done in `def header` which is what we want. The line with the test in def call` can be removed.

AdrieanKhisbe:
voilaaaaaaa :)

maxmeyer:
Wow, awesome. Thanks. :-)

AdrieanKhisbe:
\o/ :)

AdrieanKhisbe:
thanks you too @maxmeyer to have accompanied me along the way :)

PR #325, opened 2015-12-09
Move to 1.0.0

dg-ratiodata:
This PR captures all changes moving to version 1.0.0

@mattwynne @aslakhellesoy FYI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: ruby
 script: script/test
 install: script/bootstrap --without development debug
 rvm:
-  - 1.8.7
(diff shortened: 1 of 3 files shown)


mattwynne:
:tada:

mattwynne:
I don't understand why the build is failing when master is passing. Do you need to rebase some changes in?

dg-ratiodata:
Mmmh... I think rebasing with master should do the trick. Using `--debug` etc. on JVM creates huge amounts of log files which travis does not like. I fixed this on master, maybe this causes troubles here as well.

PR #332, opened 2015-12-19
Changed "Aruba" to "aruba" four times in dep warnings so you can cut/paste cleanly

jasnow:
Changed "Aruba" to "aruba" four times in dep warnings so you can cut/paste cleanly

Travis Green: https://travis-ci.org/jasnow/aruba/builds/97840424

### DIFF
--- a/lib/aruba/api/command.rb
+++ b/lib/aruba/api/command.rb
@@ -180,7 +180,7 @@ module Aruba
 
         mode = if Aruba.process
                  # rubocop:disable Metrics/LineLength
-                 Aruba.platform.deprecated('The use of "Aruba.process = <process>" and "Aruba.process.main_class" is deprecated. Use "Aruba.configure { |config| config.command_launcher = :in_process|:debug|:spawn }" and "Aruba.configure { |config| config.main_class = <klass> }" instead.')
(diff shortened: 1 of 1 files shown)


maxmeyer:
Sorry, @jasnow But I must admit I didn't checked the message you change correctly. It definitively needs to be `Aruba`, because `.configure` is a class method of the aruba module. See https://github.com/cucumber/aruba/blob/master/lib/aruba/config.rb#L85.

jasnow:
```
aruba.config.command_launcher = :in_process
aruba.config.main_class = VcrFriendlyMain
```
worked for https://github.com/georgecodes/advice-grabber

so clearer message is:
 * Use "aruba.config.command_launcher = :in_process|:debug|:spawn" and "aruba.config.main_class = \<klass\> "

maxmeyer:
Sure. But that are two different things:

```
Aruba.configure {}
```

is meant for defining defaults.

```
aruba.config.<option>
```

is meant to change/access options for specific examples/tests. If you want to configure aruba you should normally use `Aruba.configure` and not `aruba.config.<option> = <value`. Does that make sense to you?

Also have a look [here](https://github.com/cucumber/aruba/blob/master/features/configuration/usage.feature). That's a more complete explanation.

PR #333, opened 2015-12-26
Fix the post install message

ryo33:
I fixed
`gem 'cucumber', ~> '1.3.20'`
to
`gem 'cucumber', '~> 1.3.20'`


dg-ratiodata:
Oh. Thanks!

PR #347, opened 2016-03-26
avoid using removed Utils class [fix #346]

### LINKED ISSUE #346: Master still uses Utils instead of Aruba.platform
## Summary

Utils isn't defined anywhere anymore:

https://github.com/cucumber/aruba/blob/3c46eecb32b9768a014b29f2b925510e8b0c9074/lib/aruba/platforms/command_monitor.rb#L105

## Possible Solution

Use `Aruba.platform.detect_ruby` instead.

e2:


maxmeyer:
Thanks a lot.

PR #348, opened 2016-03-29
stop supporting ruby 1.8

e2:
It's unsupported by Aruba anyway (see history), so it makes no sense to:

- run integration tests (Travis/AppVeyor)
- have any code specific to any version < 1.9.3

Main inconvenience: I can't even install Ruby 1.8.7 using RVM, because it requires an existing system ruby. And that's EOL also. So it's detrimental to go back and forth to tweak PRs just so the integration builds are green. (Also, hashes are just a PITA).

Also, Aruba 1.x hasn't been release, and 1.9.3 has been EOL for over a year.

And since the differences between 1.9.3 and 2.0 aren't that horrific, the master branch may as well just assume Ruby 2.0. (Especially since Aruba is not considered a production-environment gem). But, eliminating 1.8.7 code is a huge step, nevertheless.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: ruby
 script: script/test
 install: script/bootstrap --without development debug
 rvm:
-  - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
   - 2.1.8
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :debug do
     gem 'pry-byebug', '~> 3.1.0'
   end
 
-  if RUBY_VERSION < '2' && RUBY_VERSION > '1.9' && !RUBY_PLATFORM.include?('java')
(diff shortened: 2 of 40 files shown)


dg-ratiodata:
Wow! Thanks a lot.

> It's unsupported by Aruba anyway (see history), 

No, I re-introduced support for that to make newer version of `aruba` work with `rspec`. Support was implicit when I took over maintainership. With the required_ruby option in gemspec it is at least explicit. 

I would really, really, really love to merge that, but that needs to wait until 1. :-( Sorry!

But:

I'm would like to work on 1.0.0 to get rid of so much untested/old code, so I think what makes sense here is to branch off release/1.0.0 and merge your PR there. I just pushed release/1.0.0-branch. Please open a new PR referencing this one again `release/1.0.0`.

> And since the differences between 1.9.3 and 2.0 aren't that horrific, 

Since `aruba` is used by `cucumber` and `rspec` and both support `>= 1.9.3` for the time being, we are going to support this as well.

Cheers,
Dennis

PR #352, opened 2016-03-30
Correctly reset term colors (scripts)

e2:
On dark backgrounds, black was unreadable.
In most cases, e[39;39m should correctly reset the fg and bg colors.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -2,8 +2,8 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


maxmeyer:
Very good! Thanks.

e2:
Should be: 39;49m in both cases. Oh well... still better than black.

PR #354, opened 2016-03-30
stop supporting ruby 1.8

e2:
## Summary

Remove all code, checks and RuboCop enabling/disabling for specifics related to Ruby < 1.9.2.

## Motivation and Context

Ruby 1.9.3 has been EOL for almost a year, so ideally users should move to Ruby 2.x. This makes Ruby 1.8.7 just too ancient to work with. E.g. I can't even install 1.8.7 without bending over backwards (because building 1.8.7 requires another ruby present, older versions of libraries). Pretty much the only place where 1.8.7 used nowadays is on Travis, etc.

Removing this would make contributing much easier, especially without the need to go back-and-forth with forgotten old-style hashes (` :a => :b` vs `a: :b`).

## How Has This Been Tested?

Green on Travis. (Running tests on master locally still means lots of issues).

## Types of changes

(Bug, since Ruby 1.8.7 is difficult to install and there's not official support to change that).

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: ruby
 script: script/test
 install: script/bootstrap --without development debug
 rvm:
-  - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
   - 2.1.8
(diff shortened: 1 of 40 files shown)


maxmeyer:
Thanks! Your help is really appreciated! :-) New build errors can be fixed later. :)

PR #357, opened 2016-03-31
Fix broken exception handler

e2:
## Summary

An old, undefined variable (`cmd`) was left over after refactoring. It was replaced with the right one (`commandline`).

Also, a spec was added to force the code to fail on Unix.

## Details

This caused failures on AppVeyor, because the exception handler crashed.

This failed only on Windows because the PATH traversing didn't throw an error earlier.

Still, the error could've occurred on Unix if the program was in the PATH (so no exception there), and yet there was an error accessing a program's file (like if the file wasn't readable). 

## Motivation and Context

Fixes failure on AppVeyor.

## How Has This Been Tested?

Locally, RSpec.

This probably can be simulated in Cucumber scenarios by having an unreadable binary file (e.g. using `chmod -r`)

## Types of changes

## Checklist:

### DIFF
--- a/lib/aruba/processes/basic_process.rb
+++ b/lib/aruba/processes/basic_process.rb
@@ -115,16 +115,13 @@ module Aruba
       def after_run; end
 
       def inspect
-        out = stdout(:wait_for_io => 0) + stderr(:wait_for_io => 0)
+        out = truncate("#{stdout(:wait_for_io => 0).inspect}", 35)
+        err = truncate("#{stderr(:wait_for_io => 0).inspect}", 35)
 
-        out = if out.length > 76
-                out[0, 75] + ' ...'
-              else
-                out
-              end
-
-        format '#<%s:%s commandline="%s": output="%s">', self.class, self.object_id, commandline, out
+        fmt = '#<%s:%s commandline="%s": stdout=%s stderr=%s>'
+        format fmt, self.class, self.object_id, commandline, out, err
       end
+
       alias to_s inspect
 
       private
@@ -137,6 +134,11 @@ module Aruba
         return Shellwords.split(commandline)[1..-1] if Shellwords.split(commandline).size > 1
 
         []
+      end
+
+      def truncate(string, max_length)
+        return string if string.length <= max_length
+        string[0, max_length - 1] + ' ...'
       end
     end
   end
--- a/spec/aruba/spawn_process_spec.rb
+++ b/spec/aruba/spawn_process_spec.rb
@@ -56,5 +56,39 @@ RSpec.describe Aruba::Processes::SpawnProcess do
 
       it { expect {process.start}.to raise_error Aruba::LaunchError }
     end
+
+    context "with a childprocess launch error" do
+      let(:child) { instance_double(ChildProcess::AbstractProcess) }
+      let(:command) { 'foo' }
+
+      before do
+        allow(ChildProcess).to receive(:build).and_return(child)
+
+        # STEP 1: Simlulate the program exists (but will fail to run, e.g. EACCESS?)
+        allow(Aruba.platform).to receive(:which).with(command, anything).and_return("/foo")
+
+        # STEP 2: Simulate the failure on Windows
+        allow(child).to receive(:start).and_raise(ChildProcess::LaunchError, "Foobar!")
+
+        # TODO: wrap the result of ChildProcess.build with a special Aruba
+        # class , so the remaining mocks below won't be needed
+        allow(child).to receive(:leader=)
+
+        io = instance_double(ChildProcess::AbstractIO)
+        allow(io).to receive(:stdout=)
+        allow(io).to receive(:stderr=)
+
+        allow(child).to receive(:io).and_return(io)
+
+        allow(child).to receive(:duplex=)
+        allow(child).to receive(:cwd=)
+
+        allow(child).to receive(:environment).and_return({})
+      end
+
+      it "reraises LaunchError as Aruba's LaunchError" do
+        expect { process.start }.to raise_error(Aruba::LaunchError, "It tried to start foo. Foobar!")
+      end
+    end
   end
 end
--- a/lib/aruba/processes/spawn_process.rb
+++ b/lib/aruba/processes/spawn_process.rb
@@ -83,7 +83,7 @@ module Aruba
             sleep startup_wait_time
           end
         rescue ChildProcess::LaunchError => e
-          raise LaunchError, "It tried to start #{cmd}. " + e.message
+          raise LaunchError, "It tried to start #{commandline}. " + e.message
         end
 
         after_run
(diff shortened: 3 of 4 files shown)


spec/aruba/spawn_process_spec.rb
```diff
@@ -56,5 +56,35 @@
 
       it { expect {process.start}.to raise_error Aruba::LaunchError }
     end
+
+    context "with a childprocess launch error" do
+      let(:child) { instance_double(ChildProcess::AbstractProcess) }
```
dg-ratiodata: Mmhh... I think we should not mock "Classes" we own. Just make it an double, to an instance one.

lib/aruba/processes/basic_process.rb
```diff
@@ -115,7 +115,13 @@ def before_run; end
       def after_run; end
 
       def inspect
-        out = stdout(:wait_for_io => 0) + stderr(:wait_for_io => 0)
+        out = stdout(:wait_for_io => 0)
```
dg-ratiodata: Oh... Can you give me an example. This sounds weird... I would like to understand what can cause this code to fail...

lib/aruba/processes/basic_process.rb
```diff
@@ -115,7 +115,13 @@ def before_run; end
...
+        err = stderr(:wait_for_io => 0)
+
+        # Prevent crash within a crash
+        return super if out.nil? || err.nil?
+
+        out = out + err
```
dg-ratiodata: what about renaming it to `result`? Overwriting `out` looks weird.

spec/aruba/spawn_process_spec.rb
e2: ChildProcess is from the `child_process` gem, that's why I mocked it out. (https://github.com/jarib/childprocess).

e2: If it's made a double, we miss out on RSpec's "method check" feature.

lib/aruba/processes/basic_process.rb
e2: If inspect is called before there's any stdout, they're `nil`, which causes adding them (`+`) to fail.

`inspect` methods shouldn't ever fail, or you can get a chain of failures of handling failures. Very confusing in practice, especially to contributors who do not expect things like `inspect` to every fail.

e2:
I've reworked it and added specs.

spec/aruba/spawn_process_spec.rb
dg-ratiodata: I know, but don't fiddle with someone else's classes. Why not using this instead: https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles/using-a-class-double

lib/aruba/processes/basic_process.rb
```diff
@@ -115,16 +115,13 @@ def before_run; end
       def after_run; end
 
       def inspect
-        out = stdout(:wait_for_io => 0) + stderr(:wait_for_io => 0)
+        out = truncate("#{stdout(:wait_for_io => 0).inspect}", 35)
```
dg-ratiodata: Beautiful. I like that implementation much better!

lib/aruba/processes/basic_process.rb
```diff
@@ -115,16 +115,13 @@ def before_run; end
...
+        err = truncate("#{stderr(:wait_for_io => 0).inspect}", 35)
```
dg-ratiodata: 35 <> 76. Please keep it 76 characters long for now. Strictly speaking this is an API change. Why was that changed? Normally terminals should be 80 characters wide.

dg-ratiodata:
Please address my comments / fix ruby 1.8.7 - sorry for that - and we're good to go.

spec/aruba/spawn_process_spec.rb
e2: No, that won't work. Or I don't see how.

A class double would help if I was only mocking out `ChildProcess.build`. And since `Aruba::Processes::SpawnProcess` breaks the Law Of Demeter here (the result of `.build` should be wrapped by another Aruba class), I can't use a `const_double` for it. 

I mean, I can used `const_double` for `.build` because technically it "fits" (as a method on a constant), but that's not what I'm testing. If I did, I'd be forced to use `:transfer_nested_constants` because everything in the tests is namespaced with `ChildProcess`. And that wouldn't make sense, because `ChildProcess` is the ONLY case here of using a class method.

As for "fiddling with someone else's classes" - that's the point of using `instance_double` in the first place - to make the implementation based on the interface (of the 3rd party class) while being independent from the behavior. Thus, "mocking".

It may seem debatable whether using `ChildProcess::AbstractProcess` is a good idea, but I think it's the best option anyway. Even if I used `ChildProcess::Windows::Process` instead (if it was available without requiring implicitly), it would still be the exact same interface (due to inheritance). And because of inheritance, `ChildProcess::AbstractProcess` is the public API of `childprocess` anyway, so that won't change without a major bump.

All I really wanted to do is mock out the `:start` method, but due to the way the class is setup (and not wrapped by another Aruba class), the way the mocks are done here just exposes a design problem already in codebase (not wrapping the 3rd party class with something easier to mock out).

I can't use `object_double` on the child, because then the test would depend on the signature of `ChildProcess::AbstractProcess#initialize` (which *is* a private API - the only method of the class!).

So it's either terrible mocking here, or changing the Aruba API to wrap the `ChildProcess.build` result.

I can use a double, but that's not any less verbose ... and ... it's would just be way of completely ignoring the API altogether.

I can't think of a better alternative here than adding a new wrapper class in Aruba.

lib/aruba/processes/basic_process.rb
e2: It was 76 for both input + output. So I just picked 35+35=70 + six characters left for another variable in the inspect string.

> Strictly speaking this is an API change. 

It isn't - no one is relying on the output of `#inspect` to rely on any schema.

>Normally terminals should be 80 characters wide.

Yes, but the length here is completely arbitrary - it was meant to just reduce 3-pages worth of instance variable dumps during an exception (BasicProcess class just has so much inside).

See here: 1d6a0ef4b4ae98dce14e6ec17485ce11a588e10d

The previous inspect string format is too long by itself anyway: `'#<%s:%s commandline="%s": output="%s">'` -> 38 characters, meaning only 42 would be left for the output string.

38 + 76 = 114

after my changes:

46 + 35 + 35 = 116

Again, it's just to reduce 3 pages of dumped internal variables into something a little more "user friendly".

Not to mention now you at least have a chance to see part of the actual error! (Previously, by appending the errors on stdout, you wouldn't see even a single character from stderr if stdout produced more than 76 characters...).

It's just semantically wrong to concat stdout and stderr, no matter what the use case is. (They're separate streams for strong reasons).

e2:
Fixed, rebased and squashed.

e2:
Can this be merged? Or is there anything else here needed?

maxmeyer:
Well done! No. I'm going to merge this now.

PR #359, opened 2016-03-31
"Compatibly" reimplement disk_usage API and logic

e2:
## Summary

Implement `ArubaPath#blocks` by guessing when it's not available for a given filesystem.

NOTE: If the block size cannot be determined, a default of 4096 is used. (I'm aware this is different from the default of `Aruba.config.physical_block_size`, which has a default of 512 - I believe it should be 4096).

## Details

The formula is basically asking "how many blocks are needed to at least fit the file's size?".

A spec was added to test various edge values.

## Motivation and Context

If `File::Stat.blocks` is nil, guessing is a better option than nothing.

Also, "disk usage" based on "blocks reported by stat()" is unreliable because of filesystem metadata and things like compression/deduplication (e.g. BTRFS).

So since disk usage is a "guess" at best anyway, I implemented this by calculating the blocks based on the file size. Primitive filesystems (e.g FAT) are likely more conservative with blocks, so it seems reasonable.

And there are network filesystem where we'll "never know" anyway.

Futher more, network filesystems won't even tell you what the `blksize` is, so that has to be "guessed" too.

I assumed: 4096 bytes. This seems reasonable because:

1. Only exotic filesystems and ones with custom settings have different values.
2. It's a sane balance between many small files and losing too much disk space.
3. Most Windows filesystems (where it's the usually case where values aren't available) usually defaults to 4k anyway
4. People likely aren't going to be using Aruba for calculating disk space on humongous filesystems
5. Most large-blocked local filesystems usually will likely be advanced enough to provide the info available anyway
6. Systems with smaller block sizes will be "overestimated", which is still a good thing. This doesn't matter much, because either the usage will exceed the device size (or be close), so people will notice ... or ... for lots of large files filling up the media, the total "margin of error" will be small.
7. Even mission-critical systems aren't likely going to rely on disk usage, which is to error-prone in theory alone.

So I see no downside. 

## How Has This Been Tested?

Specs added and they pass (locally). File::Stat is mocked, so whatever my own filesystem reports is irrelevant. 

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

(The "guessing" could be unexpected, but assuming disk usage measurement can be accurate is ... incorrect anyway).

## Checklist:

- [X] My change requires a change to the documentation.
- [X] Documentation has been updated

### DIFF
--- a/features/api/filesystem/disk_usage.feature
+++ b/features/api/filesystem/disk_usage.feature
@@ -1,14 +1,28 @@
 @unsupported-on-ruby-older-19
 Feature: Report disk usage
 
-  Sometimes you need to check, what amount of disk space a file consumes. We do
-  NOT support "directories" with `#disk_usage`. This does not work reliably
-  over different systems. Here can help `'#disk_usage`. But be careful, by
-  default it uses a block size of "512" (physical block size) to calculate the
-  usage. You may need to adjust this by using `Aruba.configure { |config|
-  config.physical_block_size = 4_096 }`. Don't get confused, if you check the
-  block size by using `File::Stat` (in ruby). It reports the block size of your
-  filesystem, which can be "4_096" for example.
+  Sometimes you need to check, what amount of disk space a file consumes.
+
+  This is useful, since many tiny files will usually take up a lot more disk
+  space than their sizes suggest.
+
+  While not be very accurate, `'#disk_usage` helps estimate this "real"
+  allocated size on disk.
+
+  NOTE 1: currently "directories" are NOT supported.
+
+  NOTE 2: Aruba assumes the 'physical block size' is 512 bytes. (It usually is).
+
+  If your OS/filesystem doesn't report the number of blocks used (and your
+  allocation unit size is not 4096 bytes), you can change the default "block
+  size" Aruba uses for calculations:
+
+  E.g. for a 32kB "allocation unit size" and 16 blocks used for a tiny file
+  (check `File::Stat.blocks` reported by Ruby), just divide the
+  two numbers to get the "physical block size" you need to set:
+
+  `Aruba.configure { |config| config.physical_block_size = 32_768 / 16 }`.
+
 
   We're gonna use the (correct) IEC-notation
   (https://en.wikipedia.org/wiki/Binary_prefix) here:
@@ -16,6 +30,7 @@ Feature: Report disk usage
     \* kilo => kibi
     \* mega => mebi
     \* giga => gibi
+
 
   Background:
     Given I use a fixture named "cli-app"
--- a/lib/aruba/api/filesystem.rb
+++ b/lib/aruba/api/filesystem.rb
@@ -372,10 +372,25 @@ module Aruba
       #
       # @result [FileSize]
       #   Bytes on disk
+
+      TYPICAL_FS_UNIT = 4096 # Very typical, except for giant or embedded filesystems
+      TYPICAL_DEV_BSIZE = 512 # Google DEV_BSIZE for more info
+
       def disk_usage(*paths)
         expect(paths.flatten).to Aruba::Matchers.all be_an_existing_path
+        expanded = paths.flatten.map { |p| ArubaPath.new(expand_path(p)) }
 
-        Aruba.platform.determine_disk_usage paths.flatten.map { |p| ArubaPath.new(expand_path(p)) }, aruba.config.physical_block_size
+        # TODO: change the API so that you could set something like
+        # aruba.config.fs_allocation_unit_size directly
+
+        block_multiplier = TYPICAL_FS_UNIT / TYPICAL_DEV_BSIZE
+        fs_unit_size = aruba.config.physical_block_size * block_multiplier
+
+        # TODO: the size argument here is unnecessary - ArubaPath should decide
+        # what the disk usage of a file is (even if Aruba.config needs to be
+        # read)
+        deprecated_block_count = fs_unit_size / block_multiplier
+        Aruba.platform.determine_disk_usage(expanded, deprecated_block_count)
       end
 
       # Get size of file
--- a/lib/aruba/aruba_path.rb
+++ b/lib/aruba/aruba_path.rb
@@ -126,8 +126,61 @@ module Aruba
     #
     # @return [Integer]
     #   The count of blocks on disk
+    #
+    # @deprecated
     def blocks
-      File::Stat.new(to_s).blocks
+      min_bytes_used = minimum_disk_space_used
+      min_bytes_used / Aruba.config.physical_block_size
+    end
+
+    # TODO: Aruba.config.physical_block_size could be allowed to be nil
+    # (So the unit size can be autodetected)
+
+    TYPICAL_FS_UNIT = 4096
+    TYPICAL_DEV_BSIZE = 512 # Google DEV_BSIZE for more info
+
+    # Report minimum disk space used
+    #
+    # This estimates the minimum bytes allocated by the path.
+    #
+    # E.g. a 1-byte text file on a typical EXT-3 filesystem takes up 4096 bytes
+    # (could be more if it was truncated or less for sparse files).
+    #
+    # Both `File::Stat` and the `stat()` system call will report 8 `blocks`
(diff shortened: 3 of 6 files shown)


e2:
Ok, turns out Ruby's documentation is misleading. (Ruby uses info provided by `stat()` directly).

blocks is **always** the number of 512 byte blocks allocated. It's always 512 bytes.

`blksize` is the "recommended size of IO", not the filesystem's allocation unit. It's "recommended" which means it's close to what's called "clusters" in the Windows world. So this value is useless if `blocks` are given, and it's probably absent anyway if `blocks` are `nil`.

And so, the `Aruba.physical_block_size` is simply misleading. It's not an allocation unit, so it isn't relevant for calculating.

In fact, it should be renamed to "fs_allocation_unit_size" and it should be 4096 by default. This is because we're calculating "disk usage". Almost no filesystem has 512 as the "block size", but almost every stat() implementation relies on `blocks` being the amount of allocated 512-sized blocks.

What you really want to know is: "what's the minimum sized block for an empty or 1-byte file?".

On Linux, this is: `blocks` * 512. That's why a 1-byte file usually has "8", so it's actually taking up 8*512=4096, which is the "allocation unit" of 4096.

This means `ArubaPath.blocks` should always return `File::Stat.blocks * 8 / Aruba.fs_allocation_unit_size`. Except that `File::Stat.blksize` is a more reliable way of discovering the `fs_allocaiton_unit_size` for any given **specific** file.

(I'm not sure about network systems).

I think the whole disk_usage API has to be redesigned. It's based on `File::Stat` values, which is the most misleading thing ever.

e2:
The commits and commits should be self-explanatory.

The biggest paradigm shift is burying the concept of `blocks` from `File::Stat`.

Instead, the number of `blocks` should be immediately converted into a "filesystem unit".

That's the difference: confusing "blocks" vs understandable "units".

Explanation:

On a tiny file, Stat.blocks returns 8, which is 8*512 bytes, which is 4096.

A file can't take up 1,2,3,4,5,6 or 7 "blocks" on my filesystem. Only 8, 16, 32, etc.
Instead, files can take up 1,2,3,... "units" of 4kB.

The changes are 100% compatible (except for the extra method to show total bytes allocated - which superseeds the method returning "blocks").

Anyway, the rest is in the PR.

dg-ratiodata:
Reading all your comments through, I think adding this functionality really was a mistake of mine:

* It's unreliable
* It works different on different combinations of OS AND FS

I really ask myself, why not deprecating and removing it. It doesn't look like it's worth to keep. It's far to complicated and hard to understand. WDYT? Sorry, you really did a good job by improving this functionality.

e2:
It's part of the API and you can't remove it until 1.x. 

So might as well have it merged to fix the issue (on AppVeyor at least) and consider dropping it from 1.x.

> It's unreliable

In terms of code, probably not. In terms of the idea of being able to estimate disk usage - it's as good as it get. 

There's `du` in the coreutils (and on Cygwin/Ming) but using it would be a lot trickier with no major upside. I'm not sure about use cases, but if there is a use case for checking disk usage, it's good enough.

> It works different on different combinations of OS AND FS

Yes and not - if the data isn't provided, estimates are quite good. The worst case is if you have:
1. Many 1-bytes files everywhere ...
2. ... which are truncated huge files ...
3. ... on a log-based FS (probably).

The issues are not much worse than with file permissions/chmod.

So by dropping this, you should also consider dropping chmod. It's unlikely to be useful across platforms, except an API like:
- "make_readonly", "make_writable"
- "make_executable", "make not executable" (POSIX)

Everything else (groups, etc.) means multi-user, and you probably can't switch users without admin access, which implies running Aruba as root, which implies and exotic setup that isn't portable.

If this is dropped due to complexity, same thing should happen with chmod. Even if they're both moved to aruba-contrib or something.

features/api/filesystem/disk_usage.feature
```diff
@@ -1,14 +1,45 @@
...
+  the `stat()` system call will report 8 `blocks` of 512 bytes, which is
+  exactly 4096 bytes. So ignore `blocks` values - unless you multiply them by
+  512 (almost always).
+
+  Currently Aruba allows you to set what's called the `physical_block_size`,
+  which is a bit misleading - it's the "512" value. So if you somehow have a
+  "filesystem unit size" of 8192 (instead of a typical 4KB), set the
+  `physical_block_size` to 1024 (yes, divide by 8: 8192/8 = 1024).
+
+  Ideally, in the future Aruba could provide `Aruba.config.fs_allocation_unit`,
```
dg-ratiodata: Let's don't add this here. This might confuse users.

lib/aruba/aruba_path.rb
```diff
@@ -120,14 +120,47 @@ def [](index)
...
+    TYPICAL_FS_UNIT = 4096
+    TYPICAL_DEV_BSIZE = 512 # Google DEV_BSIZE for more info
+
     # Report count of blocks allocated on disk
     #
     # This reports the amount of blocks which are allocated by the path.
     #
     # @return [Integer]
     #   The count of blocks on disk
+    # TODO: deprecate this method
```
dg-ratiodata: Add `@deprecated` here.

lib/aruba/aruba_path.rb
```diff
@@ -120,14 +120,47 @@ def [](index)
...
+    end
+
+    # Report minimum disk space used
+    #
+    # This reports the minimum bytes allocated by the path.
+    #
+    # @return [Integer]
+    #   Total bytes allocate
+    # TODO: deprecate this method
+    # TODO: this is recommended over the above "blocks"
```
dg-ratiodata: Add `@deprecated` here.

dg-ratiodata:
> It's part of the API and you can't remove it until 1.x. 

Sure. Deprecate and then remove.

> So by dropping this, you should also consider dropping chmod. 

Mmh... Personally I use this a lot. From my understanding it generally works, but on Windows only for a specific set of permissions, e.g. `644` etc. I would "solve" that via documentation. Let's have a separate issue for API-cleanup.

Please address my comments and we're good to go.

e2:
I reworked the comments a bit. 

I also removed the deprecation from the new method I introduced. (I can't judge if the feature is useful or not).

Technically, blocks are a more reliable way to determine the used space on Linux. On Windows, the filesystem's "cluster size" (filesystem allocation unit) is the only thing you can really use. So it would be trading accuracy for portability. Not to mention it's less confusing on the surface.

(E.g. A small EXT2 partition typically has 1kB blocks).

If you want to keep it, I'd make Aruba's configuration value `nil` by default, and use `File::Stat.blksize` whenever possible. This would make things correct on directory trees spanning multiple filesystems with different allocation unit sizes.

Anyway, this should be a separate gem.

e2:
Can this be merged?

lib/aruba/api/filesystem.rb
```diff
@@ -372,10 +372,25 @@ def with_file_content(file, &block)
       #
       # @result [FileSize]
       #   Bytes on disk
+
+      TYPICAL_FS_UNIT = 4096 # Very typical, except for giant or embedded filesystems
```
maxmeyer: Are those constants used somewhere else? If not, let's make them just variables. I find that easier to maintain.

lib/aruba/aruba_path.rb
```diff
@@ -126,8 +126,61 @@ def [](index)
...
-      File::Stat.new(to_s).blocks
+      min_bytes_used = minimum_disk_space_used
+      min_bytes_used / Aruba.config.physical_block_size
+    end
+
+    # TODO: Aruba.config.physical_block_size could be allowed to be nil
+    # (So the unit size can be autodetected)
+
+    TYPICAL_FS_UNIT = 4096
+    TYPICAL_DEV_BSIZE = 512 # Google DEV_BSIZE for more info
```
maxmeyer: Please make them variables if not use somewhere else.

maxmeyer:
Just replace the constants by variables and I'm going to merge this.

maxmeyer:
Ok. Did it myself to reduce the "backlog".

e2:
Thanks.

Issue #363, opened 2016-04-01
Improve branch-model


dg-ratiodata:
## Summary

Improve branch-model to make contributing easier.

## Expected Behavior

Taken from [here](https://gitter.im/cucumber/aruba?at=56fd558f76b6f9de194cefef).

it's mostly bugfixes and new features (like Docker), so there shouldn't be any conflicts. If that's an issue, it's best to switch master to 1.x, branch off 0.14 for backporting, and just focus on releasing 1.x ASAP. It's no problem to then work on 2.x a month from now - no one will complain because of "too many major releases". Aruba is complex, so learning is expected. And if learning is reflected in "multiple major releases", that's just an indication of how much learning (or rework) was needed, nothing else. I'm guessing it's actually more useful to just skip to 2.x (master - without releasing), branch off 1.x as unreleased as "work in progress" (too keep the branches). E.g. I'm (as a contributor) only interested in master and not backward compatibility or deprecations - so it makes no sense to add an artificial "burden" on master. You can always work on a "transition release" - but only if there's a genuine need in the community. Master branch should be optimized for quickly accepting PRs. (If PRs don't have priority over other work, contributing feels very discouraging - as if there's a ton of "bureaucracy"). It doesn't make sense to expect contributors to make backports of their own fixes (that they don't need themselves).

## Current Behavior

It takes a long time to release a major release.

## Possible Solution

Release faster.

e2:
:+1:

e2:
Release 0.1.0 was in ... 2010. So Aruba hasn't reach 1.x in more than 6 years...

e2:
I have a suggestion for this.

#### Symptoms

- Hard to accept pull requests on master. 
- Separate branch for unreleased 1.x .
- PRs tagged "future"

#### Cause

Master branch is expected to be available for 0.14.x patches.

#### Root cause

1.x development hasn't even BEGUN!

#### Why is this the root cause?

1.x has diverged too much from 0.14.x - and yet master STILL hasn't switched to 1.x.

If work on 1.x had begun, master would effectively be 1.x

#### Why the "root cause" a problem?

Because the default merge target of pull requests is master. If the master branch is currently "used as" 0.14.x, which means that, the way things are set up ...

... it's **harder** for people to contribute to 1.x because it's officially still an "experimental branch".

#### Why is the root cause still present?

Because no decision has been made to "officially" start working on 1.x. 1.x is still an "experiment" and not a development branch.

[...]

PR #385, opened 2016-04-28
Drop unsupported rubies

maxmeyer:
## Summary

This PR drops supported for unsupported rubies from 1.0.0 on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: ruby
 script: script/test
 install: script/bootstrap --without development debug
 rvm:
-  - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
   - 2.1.8
@@ -13,7 +11,6 @@ rvm:
   - jruby
   - jruby-20mode
   - jruby-21mode
-  - rbx
 matrix:
   include:
     - rvm: jruby-9.0.1.0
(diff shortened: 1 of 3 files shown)


mattwynne:
Did you run this past @myronmarston? Those of us maintaining test frameworks like to be able to support our users on older Rubies. I think this will mean RSpec and Cucumber won't be able to use Aruba 1.0.

myronmarston:
> I think this will mean RSpec and Cucumber won't be able to use Aruba 1.0.

That's true if we wanted to use Aruba 1.0 in RSpec's test suite for RSpec 3.x.  But in RSpec 4 we'll be dropping old ruby versions, too (probably 1.8.7 and 1.9.2 like you've done here) so we can just wait to upgrade to Aruba 1.0 until that point.  The old version of aruba we are on is meeting our needs just fine :).

mattwynne:
Sweet!

Ref #363

maxmeyer:
I dropped support for rubies < 1.9.3. Cucumber 2 requires > 1.9.3, correct?

I had in mind that Myron wants to drop old ruby support in the upcoming major release and that he's "fine" with version dropping in major versions in aruba.

I think, the 1.0.0 release will be a major clean up + repair release. I really want to get rid of the deprecated code which is still part of aruba for months, sometimes even years.



Am 29. April 2016 20:16:36 MESZ, schrieb Matt Wynne <notifications@github.com>:

PR #399, opened 2016-05-09
Prevent duplicate output for commands.

### LINKED ISSUE #374: Prevent duplicating stderr/stdout for `run_simple` commands
## Summary

When stderr+stdout announcing is enabled for `run_simple` in RSpec, both announcements are duplicated.

E.g. if a cli command outputs "hello world" in STDOUT, you'll see:

```
<<-STDOUT
hello world
STDOUT

<<-STDOUT
hello world
STDOUT
```
## Expected Behavior

If the command is only run once, it should print each stderr/stdout announcement only once.
## Current Behavior

Currently, the `stop()` call in `run_simple()` sends the notifications to announce the process's stderr+stdout. Then, `stop_all_commands` does the same.
## Possible Solution

The sanest "workaround" for now is: just temporarily disable announcing during the `stop` in run_simple. Since the output is gathered when the process ends anyway, this isn't a problem.

I implemented this in my own fork here:

https://github.com/e2/aruba/commit/c540485da24c71c1ad6fc3960b6211175e169943

Ideally, there should be a way to "consume" stderr/stdout during announcements. And ideally there would be a "streaming" of stderr/stdout live while the app is running.
1. Aruba + RSpec
2. Activate announcing for `stdout`
3. `run_simple` with a simple 'hello world' script
4. You'll have the announcement appear twice (once for `stop` inside `run_simple`, and once for `stop_all_commands` when the test is done)
## Context

I want to see the output of commands - and it's annoying to get errors showing up multiple times on the screen. Especially annoying is having two sets of stdout+stderr. No to mention it's very confusing. (At first I though somehow the test ran twice).
## Your Environment

Project without cucumber (RSpec only, standard setup with `aruba init --test-framework=rspec`).

dg-ratiodata:
## Summary

This should fix  #374.

## Details

If a user decides to use `run_simple`, a command is stopped twice:
1. After the configured wait time
2. On the end of test suite via `terminate_all_commands`

This makes the output appear twice. This is fixed by this PR.

It also makes a method call fail if:

1. Command already stopped, `#stop` is called again
2. Command has not been started, but is `#stop`ped

TODO: This PR needs some feature tests + Unit tests for the changed functionality.

## Motivation and Context

Having that many output makes it difficult to find information within the output.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/lib/aruba/api/command.rb
+++ b/lib/aruba/api/command.rb
@@ -70,7 +70,7 @@ module Aruba
                  all_commands
                end
 
-        cmds.each(&:stop)
+        cmds.each { |cmd| cmd.stop unless cmd.stopped? }
 
         self
       end
--- a/lib/aruba/command.rb
+++ b/lib/aruba/command.rb
@@ -48,6 +48,8 @@ module Aruba
 
     # Stop command
     def stop(*)
+      fail Aruba::UserError, 'Please start a command first, before stopping it' if __getobj__.stopped?
+
       __getobj__.stop
       event_bus.notify Events::CommandStopped.new(self)
 
@@ -56,6 +58,8 @@ module Aruba
 
     # Terminate command
     def terminate(*)
+      return if __getobj__.stopped?
+
       __getobj__.terminate
       event_bus.notify Events::CommandStopped.new(self)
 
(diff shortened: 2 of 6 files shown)


mvz:
I'm going to rebase this and see what happens.

mvz:
The change to `Command#stop` is very intrusive for now. We may revisit this later. I'm going to open a new PR that just changes behavior of `Command#terminate`.

PR #417, opened 2016-10-05
JRuby 9k fails cuke tests where the output contain non-English characters

stamhankar999:
## Summary

JRuby 9.0.5.0 fails cuke tests where the output contain non-English characters.
## Details

I tracked this down to an encoding error reported in childprocess when pumping data from the invoked command to the stdout tempfile. My fix is to set the stdout and stderr output files to be in binmode, and then the string matchers for the output declare that the actual result (read from the tempfiles as binary data) is actually UTF-8. This causes no harm in old JRuby or MRI afaict; my cuke tests for my application pass with this fix in place just like they passed before.
## Motivation and Context

Cuke tests for the DataStax Ruby driver failed against jruby 9k, a platform we are trying to support now.
## How Has This Been Tested?

Sorry, don't have aruba tests. :(

The cuke test that failed (there are several that failed the same way, but I focused on this one): https://github.com/datastax/ruby-driver/blob/master/features/asynchronous_io/executing_queries.feature#L45-L62

The error occurred in the "When it is executed" part:
https://datastax-oss.atlassian.net/browse/RUBY-163

Fixing SpawnProcess.start got past this error, but the "Then" clause started failing with an encoding error. I introduced the force_encoding of actual results in relevant functions to fix that. 

I did test that my cuke tests pass on MRI 2.3 and JRuby 1.7.x with these changes, so it doesn't seem like I introduced a regression.
## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ?] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  I'm not sure if this will break anyone. Changing those matchers to assume 'actual' is utf-8 _might_ be overzealous.
## Checklist:

### DIFF
--- a/lib/aruba/matchers/string/include_output_string.rb
+++ b/lib/aruba/matchers/string/include_output_string.rb
@@ -19,6 +19,7 @@
 #     end
 RSpec::Matchers.define :include_output_string do |expected|
   match do |actual|
+    actual.force_encoding('UTF-8')
     @expected = Regexp.new(Regexp.escape(sanitize_text(expected.to_s)), Regexp::MULTILINE)
     @actual   = sanitize_text(actual)
 
--- a/lib/aruba/matchers/string/match_output_string.rb
+++ b/lib/aruba/matchers/string/match_output_string.rb
@@ -19,6 +19,7 @@
 #     end
 RSpec::Matchers.define :match_output_string do |expected|
   match do |actual|
+    actual.force_encoding('UTF-8')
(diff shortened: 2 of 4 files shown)


stamhankar999:
FYI, I'm running aruba 0.13.0, but I tried upgrading to a newer aruba before digging deep into this and my cukes failed then, too.

maxmeyer:
@enkessler This might help a bit. This seems to be related to the encoding issue.

maxmeyer:
@stamhankar999 Sorry for the very late reply. If #434 is finished can you please rebase?

PR #433, opened 2017-06-09
Actually fail the build if rake test fails

mvz:
## Summary

Fixes script/test so its exit code is that of rake rather than of echo.

## Motivation and Context

The Travis builds seem green, but if you look at the logs there are a lot of failed scenarios. It seems a lot of things are broken and actually failing the build is a first step to fixing them.

### DIFF
--- a/script/test
+++ b/script/test
@@ -6,5 +6,4 @@ if [ "$RUN_IN_DOCKER" == '1' ]; then
   bundle exec rake "docker:run[$*]"
 else
   bundle exec rake test $*
-  echo bundle exec rake test $*
 fi


junaruga:
Looks good for me. We have to look the fact :)
The error causes from Rubocop test.

junaruga:
Adding `set -e` in the `script/test` to prevent unintended success.

Or without adding `set -e`, show exit status expressly like this.

```
diff --git a/script/test b/script/test
index bdad103..b85fcaa 100755
--- a/script/test
+++ b/script/test
@@ -2,9 +2,12 @@
 
 : ${RUN_IN_DOCKER:='0'}
 
+STATUS=0
 if [ "$RUN_IN_DOCKER" == '1' ]; then
   bundle exec rake "docker:run[$*]"
+  STATUS="${?}"
 else
   bundle exec rake test $*
-  echo bundle exec rake test $*
+  STATUS="${?}"
 fi
+exit "${STATUS}"
```

By the way, as the command line tool to do lint a bash shell script code such as Rubocop for Ruby.
I can recommend  `bashate` and `shellcheck`

https://github.com/openstack-dev/bashate
https://github.com/koalaman/shellcheck

mvz:
> The error causes from Rubocop test.

The cukes are also failing. I am preparing another PR to fix those.

mvz:
Something weird is going on in licence_finder 😕

maxmeyer:
Thanks.

Issue #405, opened 2016-06-05
Content of environment variable set by another ruby library gets lost


maxmeyer:
## Summary

The [middleman/middleman] project supports "stylus"-templates. This is a javascript library which is bundled into a ruby gem. The gem modifies the content of some nodejs environment variables which are required to be modified so that, "stylus" get be found.

## Expected Behavior

The "stylus"-gem should work with aruba.

## Current Behavior

The content of the environment variables gets lost because of aruba's isolation mechanism for environment variables.

```
```
## Possible Solution

Fix the isolation mechanism

## Steps to Reproduce (for bugs)

Try to migrate "middleman" to new version of aruba. Some tests will fail. See https://github.com/maxmeyer/middleman/commits/feature/upgrade-aruba-012 for this.
## Your Environment

- Version used: latest
- Operating System and version:
- Link to your project: https://github.com/maxmeyer/middleman/commits/feature/upgrade-aruba-012

maxmeyer:
This might be fixed by #442

mvz:
I can't reproduce this with the branch given. @maxmeyer can you?

There are commits in 0.12.0 that seem to address this (specifically: bc0f98be6d68f7ef6119a5a0f6849601b58d5faa), which is confusing since this bug applies to 0.12.0.

maxmeyer:
I'm really sorry, but I don't have time to dig into code.

mvz:
Ok, I'll look into this one then.

mvz:
This may very well be related to #608.

mvz:
Closing for now as no longer reproducible.

PR #434, opened 2017-06-09
Fix build on Travis

mvz:
## Summary

This is a follow-up of #433. It should fix all the failing scenarios that are now properly detected by Travis.

## Details

This change:
* Ensures a failed build is detected as such (at lease as far as the cucumber features are concerned).
* Fixes the build failures that were due to renaming of `cli` to `aruba-test-cli` in all the features (
* Fixes the `cucumber/setup.feature` feature that was broken from the start because it didn't require the needed part of Aruba.
* Fixes the configuration features by requiring the necessary part of Aruba. This was broken because the features replaced the default file.

## Motivation and Context

Builds were passing even though the cucumbers were failing. This made the Travis CI results effectively useless. Rather than merging in just the PR that makes the build break (#433) , this makes it pass so there is a solid base for future development.


mvz:
Additionally, this adds the 'debug' group to `licence_finder`'s ignore list, since that group is not installed on Travis CI.

mvz:
Left to do:
* [ ] Fix failing scenario on JRuby
* [ ] Speed up cukes so they stay under the 50m time limit.

mvz:
The failing scenario seems to have something to do with the combination of childprocess en non-ascii text.

mvz:
In fact, the JRuby failure looks like it's the same issue as #417.

maxmeyer:
@mvz Thanks a lot. Would you mind to rebase? I oversaw your PR and fix the `aruba-test-cli`-isse myself.

lib/aruba/processes/spawn_process.rb
```diff
@@ -267,7 +270,7 @@ def read_temporary_output_file(file)
         data = file.read
         file.close
 
-        data
+        data.force_encoding('UTF-8')
```
maxmeyer: @enkessler There seem to be some encoding issues with childprocess. Can you help us with that?

mvz:
@maxmeyer No problem, I rebased it.

lib/aruba/processes/spawn_process.rb
enkessler: I can try.

mvz:
Some new build issue has appeared to to the rebase. I'll take a look.

mvz:
@maxmeyer I'm a little confused: Are the deprecated functions supposed to be available in the upcoming version 1.0.0 or not?

maxmeyer:
> Are the deprecated functions supposed to be available in the upcoming version 1.0.0 or not?

@mvz Yes, they're all done when we release 1.0.0. But this should be done in a separate PR

features/02_configure_aruba/activate_announcer_on_command_failure.feature
```diff
@@ -10,6 +10,8 @@ Feature: Configure announcer activation on command failure
   Scenario: Default value
     Given a file named "features/support/aruba.rb" with:
```
maxmeyer: Let's rename the file to say `aruba_config.rb` instead of adding `require ...`. This is not the thing we want to show the user in this example. The require is done in the fixture in `aruba.rb`.

features/02_configure_aruba/activate_announcer_on_command_failure.feature
```diff
@@ -23,6 +25,8 @@ Feature: Configure announcer activation on command failure
   Scenario: Modify value
     Given a file named "features/support/aruba.rb" with:
     """ruby
+    require 'aruba/cucumber'
```
maxmeyer: Same here.

features/02_configure_aruba/command_runtime_environment.feature
```diff
@@ -19,6 +19,7 @@ Feature: Define default process environment
     Given a file named "spec/environment_spec.rb" with:
     """ruby
     require 'spec_helper'
+    require 'aruba/rspec'
```
maxmeyer: Why did you add this? Do you see an error? That's should already provided by the fixture in `spec/support/aruba.rb`.

features/02_configure_aruba/command_runtime_environment.feature
```diff
@@ -40,6 +41,7 @@ Feature: Define default process environment
...
```
maxmeyer: Same here

features/02_configure_aruba/command_runtime_environment.feature
```diff
@@ -63,6 +65,7 @@ Feature: Define default process environment
...
```
maxmeyer: Same here

features/02_configure_aruba/command_runtime_environment.feature
```diff
@@ -86,6 +89,7 @@ Feature: Define default process environment
...
```
maxmeyer: Same here.

features/02_configure_aruba/command_runtime_environment.feature
```diff
@@ -109,6 +113,7 @@ Feature: Define default process environment
...
```
maxmeyer: Same here.

lib/aruba/api.rb
```diff
@@ -8,7 +8,7 @@
 require 'aruba/api/core'
 require 'aruba/api/command'
 
-if Aruba::VERSION <= '1.0.0'
+if Aruba::VERSION <= '1.1.0'
```
maxmeyer: I think you add this to make this PR not to fail the build, right?

lib/aruba/processes/spawn_process.rb
```diff
@@ -64,6 +64,9 @@ def start
         @stdout_file.sync = true
         @stderr_file.sync = true
 
+        @stdout_file.binmode
+        @stderr_file.binmode
```
maxmeyer: What happen's if this is not binmode?

spec/event_bus_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'spec_helper'
 require 'aruba/event_bus'
```
maxmeyer: Is this line really required after your change?

spec/aruba/hooks_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'spec_helper'
 require 'aruba/config'
```
maxmeyer: Can we remove this line after your change?

spec/aruba/processes/basic_process_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'spec_helper'
 require 'aruba/processes/basic_process'
```
maxmeyer: Same here.

spec/aruba/spawn_process_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'spec_helper'
 require 'aruba/processes/spawn_process'
```
maxmeyer: Same here.

spec/event_bus/name_resolver_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'spec_helper'
 require 'aruba/event_bus/name_resolver'
```
maxmeyer: Same here.

lib/aruba/processes/spawn_process.rb
maxmeyer: Ah ok. Thanks for the helpful commit message.

mvz:
> @mvz Yes, they're all done when we release 1.0.0. But this should be done in a separate PR

Ok, for that purpose I have updated the version checks to allow deprecated functions when the version is 1.0.0 for now, since otherwise everything breaks at once. The removal can then safely be finalized in that separate PR.

features/02_configure_aruba/activate_announcer_on_command_failure.feature
mvz: Yes, that makes sense.

mvz:
Since the version updated causes some new spec failures, I propose the following:
* I create a new PR with just the parts that make the specs run (and fail in one or two places).
* Next, I'm going to fix the one or two failing specs (I'm partway done with this) and make a second PR for that.
* After that, I can update this one and see which bits are still needed.

lib/aruba/api.rb
mvz: Yes, Aruba still uses and/or tests some deprecated functions at the moment.

spec/aruba/hooks_spec.rb
mvz: No, it seems it can go.

maxmeyer:
Thanks @mvz. BTW: I like your commit style. 

Would you mind to squash those commits handling the same lines of code - e.g. introduce the requires + get rid of those requires. Alternatively drop them.

I will have a detailed look at your PR tomorrow.

lib/aruba/api.rb
maxmeyer: Yes, might be. Some months ago I started getting rid of them, but somehow I managed this branch to get lost. Since I move my "working environment". But I'm quite eager to remove them as fast as possible. There are tons of (very) old code which is deprecated for ages.

maxmeyer:
> In fact, the JRuby failure looks like it's the same issue as #417.

Would be great if this is fixed with your PR. JRuby have given me a few headaches so far. I'm just on C-Ruby. :smile:

features/02_configure_aruba/command_runtime_environment.feature
mvz: No, it seems this not needed. I'm not sure what made me add it.

mvz:
@maxmeyer do you still want to have the tmp dir used depend on the suite (rspec vs cucumber)? It seems to me to depend too much on the end user's chosen set up whether that's useful (plus it breaks some of the current scenarios, which is why I'm asking).

lib/aruba/cucumber/hooks.rb
```diff
@@ -5,7 +5,7 @@
 
 if Aruba::VERSION >= '1.0.0'
   Around do |_, block|
-    Aruba.platform.with_environment(&block)
+    with_environment(&block)
```
maxmeyer: I did this change before and something was uterly broken afterwards - I was forced to revert that change. What's the reason for this change?

lib/aruba/cucumber/core.rb
```diff
@@ -1,4 +1,4 @@
-if Aruba::VERSION >= '1.0.0'
+if false && Aruba::VERSION >= '1.0.0'
```
maxmeyer: Personally I use both `rspec` and `cucumber` in projects. Sometimes I ran both and want the results to be availabe. That's the reason for this fix. Can you point me to the failure to see if there's something we can do about this?

features/05_use_rspec_matchers/command/run_too_long.feature
```diff
@@ -25,3 +25,24 @@ Feature: Check if a timeout occured during command execution
     """
     When I run `rspec`
     Then the specs should all pass
+
```
maxmeyer: This should be on its own feature file. :astonished:.

https://github.com/cucumber/aruba/blob/master/features/05_use_rspec_matchers/command/have_finished_within_time.feature

Did you rebase your PR?

spec/aruba/api_spec.rb
```diff
@@ -102,6 +102,26 @@
         expect(File.exist?(File.expand_path(@directory_path))).to be_truthy
       end
     end
+
+    describe '#cd' do
```
maxmeyer: Let's move the changes for env to a separate PR.

lib/aruba/rspec.rb
```diff
@@ -18,6 +18,12 @@
     if self.class.include? Aruba::Api
       restore_env
       setup_aruba
+
```
maxmeyer: This whole changing environment-thing in aruba is a bit of a hassle. I would love to this in a different PR, please.

spec/aruba/api_spec.rb
```diff
@@ -1071,6 +1071,48 @@ def actual_permissions
 
         expect(ENV[variable]).to eq '1'
       end
+
```
maxmeyer: I'm interested to see, if this #405 is fixed by your change. Would be quite happy if the answer is yes. Yet again, please move this to another PR.

maxmeyer:
@mvz I would like to merge this PR before we/you go on fixing more bugs. Can you please extract the commits handling env logic to a separate PR. I'm going to merge this one then.

spec/aruba/api_spec.rb
maxmeyer: Forget about my comment extracting the env thing. I just saw I already created a PR for that.

lib/aruba/cucumber/hooks.rb
mvz: The `Aruba.platform.with_environment` version clears the entire environment, including things like `PATH`, causing a lot of cukes to fail because they can't find executables, among other things.

mvz: I'll move this commit to the PR about env handling, by the way.

features/05_use_rspec_matchers/command/run_too_long.feature
mvz: Ah, that's where it went 😄. I'll remove this duplicate.

lib/aruba/cucumber/core.rb
mvz: There are several cucumber scenario's that check directories. These fail because they expect `tmp/aruba`, and not `tmp/cucumber`. This is fairly easy to change, but to me it makes more sense to let each user of aruba to set their own working directory if necessary, since there are many ways to parallelize the build, e.g., by running several instances of cucumber at the same time.

mvz:
I rebased this and isolated all env-related commits in #442. I'll hold of on adding more fixes for now so we can get some of this merged. I'm happy to create more separate PRs since this is quite a mixed bag of commits.

lib/aruba/cucumber/core.rb
maxmeyer: Ok, then let's remove the lines altogether instead of using `false`.

lib/aruba/cucumber/hooks.rb
maxmeyer: ok

maxmeyer:
@mvz Personally I would love to see all env related commits to be moved to that separate PR - if not already done - and then dropped from this PR. I think there a few left in here.

After the minor other things I just saw today, we're good to merge this one. After that I will have a look into the env PR.

mvz:
> and then dropped from this PR. I think there a few left in here.

Yes, all the commits of #442 are also in this PR. It's probably best to first merge #442 and then rebase this one, but I can separate them out if you prefer.

lib/aruba/cucumber/core.rb
mvz: Ok, I'll do that.

mvz:
I have now rebased this so it no longer includes the environment fixes that are in #442.

maxmeyer:
@mvz I'm going to merge this PR first. But please do not wonder, if this happens just at the beginning of next week.

BTW: Did you see my comments about the `aruba/cucumber` vs. `aruba/api`-thing? I just pressed the review finished button.

mvz:
> Did you see my comments about the aruba/cucumber vs. aruba/api-thing?

Ah yes, they only showed up just now. I'll take a look.

features/03_testing_frameworks/cucumber/setup_aruba_for_cucumber.feature
```diff
@@ -47,7 +47,7 @@ Feature: Getting started with Cucumber and aruba
 
     Given a file named "features/support/env.rb" with:
     """
-    require 'aruba/api'
+    require 'aruba/cucumber'
```
mvz: Ok, clear. I have removed that change.

mvz:
🎉

PR #439, opened 2017-06-29
Use new proposed structure for gems by bundler

maxmeyer:
## Summary

When creating new gems with bundler it uses `exec` for executable files and `bin` for helper executables local to the project.

## Types of changes

## Checklist:


mvz:
I thought it used `exe`, not `exec`? (I don't really see the point of this new schema by the way.)

olleolleolle:
@mvz One reason I can see would be people wanting `bin/` to contain only [binstubs](http://bundler.io/v1.15/man/bundle-binstubs.1.html) and such. If that were the case, all of `bin/` could be git-ignored. If that's ignored, then `exe` could contain the executables that are to be exposed to users (as annotated in the gemspec).

Your mileage may vary.

mvz:
Ah, so binstubs put stuff in `bin/` instead of `script/`, so stuff from `script/` was moved to `bin/`, causing stuff from `bin/` to need to move to `exe/`. That kind of makes sense.

PR #442, opened 2017-06-30
Fix environment handling

mvz:
## Summary

Improve handling of environment. In particular, makes `ENV['HOME']` have a correct and safe value.

~~This PR depends on #441 so needs to be rebased after that PR has been merged.~~

## Motivation and Details

The main purpose is to make all the current specs pass. It does this by changing the environment setup in the specs, so that:

* HOME and PATH are set early enough so that all specs run with the set values. This also ensures no leaking of files outside of the build directory.
* Where the Aruba runtime is the object under test, it uses the test object when expanding paths.

This change also fixes the `#cd` method to not clear the entire environment inside the block.

After this PR has been merged we can revisit #434 and concentrate on the cucumber scenarios there.

### DIFF
--- a/lib/aruba/api/core.rb
+++ b/lib/aruba/api/core.rb
@@ -61,7 +61,7 @@ module Aruba
 
             Aruba.platform.chdir File.join(aruba.root_directory, aruba.current_directory)
 
-            result = Aruba.platform.with_environment(
+            result = with_environment(
               'OLDPWD' => old_dir,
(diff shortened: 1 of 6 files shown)


maxmeyer:
@mvz I forgot why I implemented it with `Aruba.platforms.with_environment`. At that time I tried to upgrade `cucumber` and/or `rspec` to use the new version of `aruba`. But anyway I would like to see, where this leads to. At least if fixes a few things, and if something is broken afterwards we can fix it.

mvz:
🎉

PR #447, opened 2017-07-10
Rakefile: wrap in bundle exec

olleolleolle:
- I am trying to make the Travis build run license_finder

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,17 +34,17 @@ end
 namespace :test do
   desc 'Run cucumber tests'
   task :cucumber do
-    sh 'cucumber'
+    sh 'bundle exec cucumber'
   end
 
   desc 'Run cucumber tests which are "WORK IN PROGRESS" and are allowed to fail'
   task :cucumber_wip do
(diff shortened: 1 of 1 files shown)


olleolleolle:
@maxmeyer Oh, my investigation was not final - and the license_finder still fails in Travis, trying to locate the right gems in the Gemfile.lock - https://travis-ci.org/cucumber/aruba/jobs/252143170#L340 is an example of that.

maxmeyer:
There's another PR fixing a lot issues from @mvc.

olleolleolle:
@maxmeyer Thanks for linking to that! It explains so much

mvz:
This commit shouldn't be needed since AIUI, all `bundle exec` does is set some environment variables, and these will already be inherited from the outer level `bundle exec rake`.

maxmeyer:
@mvc  You're right, I think that was the reason, why I didn't add it. but it doesn't do any harm for those using rake directly, right?

mvz:
No, I don't think it does any harm.

PR #450, opened 2017-07-12
Split up API spec files

maxmeyer:
## Summary

This PR cleans up the `spec`-directory - Focus: `spec/aruba/api`.

## Motivation and Context

* `spec/aruba/api_spec.rb is ~1000 lines long
* The file is hard to maintain
* To remove the deprecated methods all specs need to be removed as well, this should be easier if they are not all over one file.

## Types of changes

- [x] Refactoring

## Checklist:


mvz:
Won't it make more sense to split up these specs according to where the methods are defined? E.g., `#absolute_path` is defined in `lib/aruba/api/deprecated.rb`, so its specs would go in `spec/aruba/api/deprecated_spec.rb`.

maxmeyer:
@mvz Sounds reasable, but my idea was to just break apart the `api_spec.rb`. Then start another PR where we start removing deprecated code and just remove the test file which belongs to the code.

Issue #432, opened 2017-04-20
Regression: `exit status` not waiting for `run interactively` process to complete


mroth:
## Summary

In version `0.6.2`, a `run interactively` step followed by a `the exit status should be` step would complete, allowing the results of the process to be examined (as well as ensuring any side effects were complete before moving on to additional steps).

Upon updating to latest stable `0.14.2`, this no longer happens, causing commands to run out of order, and all the workarounds I can think of seem fairly hacky and undesirable. It appears `run interactively` is now functioning as a "background" process by default?

I believe this may be a regression/bug?, since that doesn't appear to be the goal of the behavior from sample Features.

## Expected Behavior

There should be a way to let an interactive process complete before checking it's exit status, or another mechanism to avoid "out of order" problems.

## Current Behavior

I've annotated the actual Scenario Outline in my test cases here to illustrate what is happening, see comments inline:
~~~gherkin
Scenario Outline: Wrapped `git reset` can handle files with spaces properly

  Given I am in a git repository
  Given an empty file named "file with spaces.txt"
  ###AAA
  Given I successfully run `git add "file with spaces.txt"`

  ###BBB
  Given I run `<shell>` interactively
    And I type `eval "$(scmpuff init -ws)"`
    And I type "scmpuff_status"
    And I type "git reset 1"
    And I type "echo 'DEBUG: PHASE BBB'"
    And I type "exit"
  Then the exit status should be 0
  ### ^^^ this is now checking exit status of AAA, not BBB!

  # Then the output should contain "BBB"
  ### ^^^ this command however, would force aruba to wait until BBB completes (hacky option #1)

  # Then I stop the command started last
  ### ^^^ (hacky option #2) also seems to ensure BBB completes, but throws a deprecation error

  ###CCC
  ### currently, this phase is failing, because it is taking place *BEFORE* BBB completes
  When I run `scmpuff status`
  Then the stdout from "scmpuff status" should contain:
    """
    untracked:  [1] file with spaces.txt
    """
  Examples:
    | shell |
    | bash  |
    | zsh   |
~~~

## Possible Solution

I've indicated two possible workarounds in the above code sample comments, but neither seems like it's the intended solution (and one throws a deprecation warning).

## Context & Motivation

In our tests, we need to do some stuff in an interactive shell (as our program modifies the shell environment for end-users as it's primary use case), and follow that up by additional (non-interactive) tests afterwards to check on the state of things.

## Your Environment

[...]

PR #452, opened 2017-07-12
Fix bugs in features/aruba

maxmeyer:
## Summary

Fix CI errors.

## Details

This PR fixes errors on CI to make the build pass again.

## Motivation and Context

Make the build pass again.

## Types of changes

## Checklist:

### DIFF
--- a/features/04_aruba_api/core/expand_path.feature
+++ b/features/04_aruba_api/core/expand_path.feature
@@ -56,8 +56,7 @@ Feature: Expand paths with aruba
 
   Scenario: Use ~ in path
 
-    Now this useses the HOME-variable from your normal shell HOME-variable.
-    From 1.0.0 on HOME will be `File.join(aruba.config.root_directory,
+    Aruba sets HOME to `File.join(aruba.config.root_directory,
     aruba.config.working_directory)`. If you want HOME have some other value,
     you need to configure it explicitly via `Aruba.configure {}`.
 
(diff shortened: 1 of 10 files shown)


maxmeyer:
@mvz Feel free to push to this PR/branch.

features/04_aruba_api/core/expand_path.feature
```diff
@@ -69,7 +69,7 @@ Feature: Expand paths with aruba
     RSpec.describe 'Expand path', :type => :aruba do
       let(:path) { '~/path/to/dir' }
 
-      it { expect(expand_path(path)).to match %r</home/[^/]+/path/to/dir> }
+      it { expect(expand_path(path)).to match %r</home/([^/]+/)+path/to/dir> }
```
mvz: I think the intent here was for `[^/]+` to match exacly a user name.

The correct fix is probably to remove the 'before 1.0.0' part of the feature and fix the 1.0.0 part to not set things up in the `before` section.

maxmeyer: No, I think, it was a "hack" I introduced. But you're right, the correct way was to remove that test. I missed the comment last night. Thanks.

mvz:
Some of the remaining failures are due to Aruba not waiting for the running command to stop. @maxmeyer can you advise on what the desired behavior should be? I'm talking about cases where right now the cucumber scenarios have this:

```
  Scenario: Running aruba interactively

    This starts a pry console with "aruba"'s api loaded.

    When I run `bin/console` interactively
    And I type "exit"
    Then the exit status should be 0
```

The last step does not wait for any command to finish, unlike the old pre-1.0 behavior.

maxmeyer:
@mvz 

* First of all we've got an issue for this behaviour: https://github.com/cucumber/aruba/issues/432
* I thought about making stopping a command more explicit in aruba 1.0.0 or at least in 2.0.0, with a cucumber step, but I think that need's to be discussed upfront
* Personally I would expect aruba to wait a couple of seconds before stopping the command and checking the status
* This needs to work both with interactive and non-interactive commands

> The last step does not wait for any command to finish, unlike the old pre-1.0 behavior.

Unfortunately I think I broke this step "a long time ago" without having issues on CI for a while. At least the code in several parts which are involved is the same for quite a long time.

What do you think about this?

maxmeyer:
@mvz I forced pushed a rebased version of this branch. Just for you to know. I removed the `if`-statements along with the changed tests.

maxmeyer:
@mvz I would like to merge this branch and then setup another one to capture the discussion around the problems with failing tests due to aruba not waiting for the command to finish.

mvz:
@maxmeyer sounds like a good plan.

PR #455, opened 2017-07-16
Remove config.keep_ansi option

mvz:
## Summary

This removes the `keep_ansi` config option. This option has been replaced by `remove_ansi_escape_sequences` so should be removed before the 1.0.0. Keeping both led to the undesirable situation that both have to be changed from the default to actually keep the ansi escape sequences, unlike other deprecated functionality which just leads to a warning message.

This also removes deprecated methods that depended on this option.

## Types of changes

This is a breaking change that is appropriate for the 1.0.0 release.


mvz:
Ok, I'll park this PR for now then. I just don't want to forget about keep_ansi since there is no deprecation warning for old config keys.

maxmeyer:
@mvz Good!

mvz:
@maxmeyer I rebased this on current master. It no longer includes the removal of `#unescape` since that was removed in #488.

mvz:
Now part of #508

PR #465, opened 2017-07-26
Fix problems with API-calls based on last_command_started

maxmeyer:
## Summary

There are some API-calls which depend on `#last_command_started'. The behaviour changed with aruba >= 1.0.0. Now all commands need to be stopped explicitly. This PR clarifies this.

## Motivation and Context

TODO:

* [ ] Add examples for the new behaviour
* [ ] Find places where we need to stop commands explicitly

## Types of changes

- [ ] Update documentation

## Checklist:

### DIFF
--- a/lib/aruba/platforms/command_monitor.rb
+++ b/lib/aruba/platforms/command_monitor.rb
@@ -45,18 +45,7 @@ module Aruba
       raise ArgumentError, e.message
     end
 
-    if Aruba::VERSION < '1'
-      # Return the last command stopped
-      def last_command_stopped
-        return @last_command_stopped unless @last_command_stopped.nil?
-
-        registered_commands.each(&:stop)
-
-        @last_command_stopped
-      end
-    else
-      attr_reader :last_command_stopped
-    end
+    attr_reader :last_command_stopped
 
     # Set last command started
     #


maxmeyer:
@mvz I think the changed code in the first commit causes the failing tests. What do would you expect as a user? Shall we discuss this in #432?

mvz:
I wonder if this is the same problem as #432, since that was with a pre-1.0 version.

maxmeyer:
Not sure. Will have a look into this in the next couple of day's, but feel free to help.

mvz:
In terms of what I expect as a user, it is a bit tricky.
* If I'm just running one command, I expect this step to wait for my command to finish.
* If I run commands sequentially, I expect this step to wait for the command started last to finish (this is the #432 case I think).
* If I'm running two commands in parallel, I probably expect this step to wait for the command I interacted with last to finish.

As a first approximation, I think the step 'the exit status should be <n>` needs to wait for the last started command to exit. We can then see what special cases are needed for the third case above.

mvz:
I think this was fixed in #487.

maxmeyer:
Thanks @mvz for fixing this.

Issue #470, opened 2017-07-28
Reducing running platforms in Travis


junaruga:
## Summary

Considering supported platforms.

## Current Behavior

Right now aruba supports below platforms.
14 VMs (Virtual Machines) are running.
https://github.com/cucumber/aruba/blob/master/.travis.yml

Total running time is over than 1 hour.
https://travis-ci.org/cucumber/aruba/builds/258426774
> Ran for 1 hr 4 min 4 sec

I think that this situation is what we have to improve.

## Possible Solution

Cucumber supports below platforms.
https://github.com/cucumber/cucumber-ruby/blob/master/.travis.yml

I guess that we can remove below 7 cases?

```
  - 1.9.3
  - 2.0.0
  - jruby
  - jruby-20mode
  - jruby-21mode
```

```
    - rvm: jruby-9.1.12.0-20mode
      env: JRUBY_OPTS='--dev'
    - rvm: jruby-9.1.12.0-21mode
      env: JRUBY_OPTS='--dev'
```

## Context & Motivation

Reducing the number of VMs, saving the resource improves the total running time.
I have experienced similar situation in another project.
After reducing the number of VMs, the running time has been improved.

junaruga:
As information, currently All the JRuby tests are error on Travis CI because of exceeding running time.
https://travis-ci.org/cucumber/aruba/builds/258937032

In my opinion, we should run only JRuby latest stable version 9.1.12.0 for JRuby tests as allow failure.
Then let's try to do tuning the tests.

junaruga:
@maxmeyer how do you think of this topic?

olleolleolle:
I support the removal of the "variant" builds for JRuby.

maxmeyer:
## In General

@junaruga Yes, I would love to remove those outdated platforms. I like the way cucumber's `.travis.yml` is structured + they seem to test on MacOS. Let's take over their structure and add the "old" versions for now, but get rid of the `-mode`-variants.

~~~
  - 1.9.3
  - 2.0.0
  - jruby
~~~

Just for the records: We already using Travis' "Docker" infrastructure, so no real VMs are started for "aruba"'s builds.


## Remove all old versions:

Pros:
* Less work on our side
* Better response times on Travis

Cons:
* Cut support for those ruby versions in 1.0.0 would make it hard for projects like `rspec` to migrate to 1.0.0. RSpec for example even support Ruby 1.8.7.

Suggestion:
[...]

PR #476, opened 2017-07-31
Update Travis Configuration

maxmeyer:
## Summary

This is a follow up to #470. It cleans up the travis configuration.

## Details

* Remove jruby-variants
* Add Mac OS builds

## Motivation and Context

See issue #470.

## How Has This Been Tested?

Travis Build

## Types of changes

- [x] Refactoring (cleanup of codebase withouth changing any existing functionality)
- [ ] Update documentation

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,47 +2,70 @@ sudo: false
 language: ruby
 script: bin/test
 install: bin/bootstrap --path vendor/bundle --without development debug
-rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1.10
-  - 2.2.7
-  - 2.3.4
-  - 2.4.1
-  - ruby-head
-  - jruby
-  - jruby-20mode
-  - jruby-21mode
-  - jruby-9.1.12.0
 matrix:
   include:
+    - rvm: ruby-head
+      os: linux
+      # - rvm: ruby-head
+      # - os: osx
+      # # TravisCI OSX doesn't work with Ruby 2.4.1 yet (https://github.com/travis-ci/travis-ci/issues/7848)
+    - rvm: 2.4.1
+      os: linux
+      # - rvm: 2.4.1
+      # - os: osx
+      # # TravisCI OSX doesn't work with Ruby 2.4.1 yet (https://github.com/travis-ci/travis-ci/issues/7848)
+    - rvm: 2.4.0
+      os: osx
+    - rvm: 2.3.4
+      os: linux
+    - rvm: 2.3.4
+      os: osx
+    - rvm: 2.2
+      os: linux
+    - rvm: 2.2
+      os: osx
+    - rvm: 2.1
+      os: linux
+    - rvm: 2.1
+      os: osx
+    - rvm: 2.0.0
+      os: linux
+    - rvm: 2.0.0
+      os: osx
+    - rvm: 1.9.3
+      os: linux
+    - rvm: 1.9.3
+      os: osx
+    - rvm: jruby
+      os: linux
     - rvm: jruby-9.1.12.0
-      env: JRUBY_OPTS='--dev'
-    - rvm: jruby-9.1.12.0-20mode
-      env: JRUBY_OPTS='--dev'
-    - rvm: jruby-9.1.12.0-21mode
-      env: JRUBY_OPTS='--dev'
+      os: linux
+    - rvm: jruby-9.1.12.0
+      os: osx
+      env:
   allow_failures:
-  - rvm: ruby-head
-  - rvm: jruby-9.1.12.0
-    env: JRUBY_OPTS='--dev'
-  - rvm: jruby-9.1.12.0-20mode
-    env: JRUBY_OPTS='--dev'
-  - rvm: jruby-9.1.12.0-21mode
-    env: JRUBY_OPTS='--dev'
+    - rvm: ruby-head
+    - rvm: jruby
+    - rvm: jruby-9.1.12.0
   fast_finish: true
 branches:
   only:
     - master
     - still
-notifications:
-  webhooks:
-    urls:
-      - https://webhooks.gitter.im/e/d4d9080e2a8c4910f609
+# notifications:
+#   slack: cucumberbdd:oQFVhzsx4R94KWmjlejAJYnM
 env:
   global:
    - secure: l8uznA5K4K9mZ1krmP3lTMD8WcJ32qGxFOR3jubKHcOBSLB4xSzU2aIqjyJdO+rLzebkwamhJc8pGSIWOUDQYvFiX7splK+uEkbBJ5huAhXtLF4Qgl86bCWbEXYzN7rvn0DQfpJAovyFMNRMnfo70XhwqWzFsaYa7Z0YbqYsJE4=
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,47 +2,85 @@ sudo: false
...
+    - rvm: 1.9.3
+      os: osx
+    - rvm: jruby
+      os: linux
+      env:
+        # Travis by default also have "-Dcext.enabled=false" set in
+        # JRUBY_OPTS, but JRuby 9 does not support C extensions at all
+        # so it issues warning that will mess up the sterr checks.
+        - JRUBY_OPTS="--server -Xcompile.invokedynamic=false"
+        # Somehow a "ASCII-8BIT to UTF-8 conversion error" appears for
```
mvz: This would also need to be fixed. I'll open a ticket so we don't forget.

mvz: Done: https://github.com/cucumber/aruba/issues/477

.travis.yml
```diff
@@ -2,47 +2,85 @@ sudo: false
...
+      os: linux
+    - rvm: 1.9.3
+      os: osx
+    - rvm: jruby
+      os: linux
+      env:
+        # Travis by default also have "-Dcext.enabled=false" set in
+        # JRUBY_OPTS, but JRuby 9 does not support C extensions at all
+        # so it issues warning that will mess up the sterr checks.
+        - JRUBY_OPTS="--server -Xcompile.invokedynamic=false"
```
mvz: The `--server` option increases startup time, which is very relevant for Aruba's cucumber tests since they start up Aruba a lot. Let's try `--client` and see if that makes any of the JRuby targets finish in time.

mvz: I've done some experiments locally, and the following combination of environment variables greatly improves test time:

```
export JRUBY_OPTS="--client -Xcompile.invokedynamic=false"
export JAVA_OPTS='-Djruby.compile.mode=OFF -XX:+TieredCompilation -XX:TieredStopAtLevel=1'
```

I've tried this with just `bundle exec cucumber features/01_getting_started_with_aruba/run_commands.feature`, and it improved clock runtime from 1m11s to 34s, while reducing system load.

I found these options at https://gist.github.com/lucasallan/2f9b5c945280d226916c and https://github.com/jruby/jruby/wiki/Improving-startup-time.

.travis.yml
```diff
@@ -2,47 +2,85 @@ sudo: false
...
+        - LC_ALL=en_US.UTF-8
+        - LANG=en_US.UTF-8
+        - LANGUAGE=en_US.UTF-8
+    - rvm: jruby-9.1.12.0
+      os: osx
+      env:
+        # Travis by default also have "-Dcext.enabled=false" set in
+        # JRUBY_OPTS, but JRuby 9 does not support C extensions at all
+        # so it issues warning that will mess up the sterr checks.
+        - JRUBY_OPTS="--server -Xcompile.invokedynamic=false"
```
mvz: Can we move setting of these variables to a global section to avoid duplication?

maxmeyer:
Yes. That makes sense. Thanks a lot for your feedback. I will try to push a new version tomorrow evening taking into account your findings.

.travis.yml
```diff
@@ -2,47 +2,85 @@ sudo: false
...
-  - jruby
-  - jruby-20mode
-  - jruby-21mode
-  - jruby-9.1.12.0
 matrix:
   include:
+    - rvm: ruby-head
+      os: linux
+    - rvm: ruby-head
+      os: osx
```
junaruga: Seeing the result of Travis, ruby-head (osx) environment in Travis server has problem right now.
You may be able to remove the ruby-head (osx) if you do not like to get the error every time.
https://travis-ci.org/cucumber/aruba/jobs/259520443

mvz:
The java variants still don't finish in time 😞

maxmeyer:
I pushed to commits to check taking the tip from @junaruga into account by reducing the number of jobs per build.

maxmeyer:
https://travis-ci.org/cucumber/aruba/jobs/260478099 is with jruby only, but it doesn't matter how many parallel tests we're running. jruby also complains about a missing `--debug` option

maxmeyer:
According to https://github.com/jruby/jruby/wiki/Improving-startup-time#use-the---dev-flag

Those flags are set by `--dev`

* client mode where applicable (generally older 32-bit JVMs)
* TieredCompilation and TieredStopAtLevel=1, equivalent to client mode on newer Hotspot-based JVMs
* compile.mode=OFF to disable JRuby's JVM bytecode compiler
* jruby.compile.invokedynamic=false to disable the slow-to-warmup invokedynamic features of JRuby

mvz:
> Those flags are set by --dev

That makes things a lot simpler! 👍

maxmeyer:
Last build:
00:00 min - start build
06:00 min -  setup project to run tests
07:00 min - run licence finder
07:30 min - run rspec
08:30 min - run cucumber
[TBC]

mvz:
I'd like to make the following suggestions to get things moving and to avoid dealing with endless backward-compatibility problems:

* Drop support for old rubies now, before the 1.0.0 release. We can discuss what is a good definition of 'old'.
* Move the JRuby variants to the `allow_failures` section.

maxmeyer:
Currently we're not facing any backward compatible problems with any of the MRI rubies, correct? "Only" jruby is a "troublemaker". I ran the suite on my laptop with jruby and it was "slow" as well. So I think let's move the jrubies to the `allowed_failures` section - as you @mvz proposed - and try to tackle this after we fixed all other problems / polished the release. 

Am 2. September 2017 20:35:24 MESZ schrieb Matijs van Zuijlen <notifications@github.com>:

mvz:
> Currently we're not facing any backward compatible problems with any of the MRI rubies, correct?

I should have explained more. Yes, there are no problems right now, but if we do not drop _official_ support for those old rubies right now, Aruba could be stuck with them for the entire 1.x cycle, meaning we cannot use modern Ruby syntax even if we want to, and/or we have to keep version checks lying around. Finally, debugging any issues on those older Rubies may require actually installing them which will become harder and harder.

A counter-argument is the desire for RSpec to use Aruba 1.0 voiced in https://github.com/cucumber/aruba/issues/363#issuecomment-215544753.

mvz:
Moving JRuby to `allowed_failures` is definitely more urgent :-).

maxmeyer:
@mvz you're right. personally I think we could release 2.0.0 right after 1.0.0 which drops support for those rubies. Not sure how to handle support best.

mvz:
Well, right after may be a little too soon, if we agree that releasing 2.0.0 in 2018 is fine then I'm fine with keeping support for Ruby 1.9.3 in 1.x.

mvz:
@maxmeyer are you going to update this PR or would you like me to do it?

maxmeyer:
Feel free to fix it!

mvz:
I'm going to open a new PR for this to rebase and clean up the branch.

maxmeyer:
I rebased the PR based on your work and squashed everything into a single commit. Feel free to merge if you're happy with this PR.

Issue #492, opened 2017-09-07
Travis builds are failing due to a weird interaction with gem resolution by Bundler


mvz:
## Summary

I thought @maxmeyer and I fixed the builds but now they're failing because of a version conflict for `did_you_mean`.

## Expected Behavior

Bundler activates `did_you_mean` version `1.0.3` and proceeds to run the tests.

## Current Behavior

`bin/test` fails with the following error (from https://travis-ci.org/cucumber/aruba/jobs/272342270):

```
$ bin/test
bundler: failed to load command: rake (/home/travis/build/cucumber/aruba/vendor/bundle/ruby/2.3.0/bin/rake)
Gem::LoadError: You have already activated did_you_mean 1.0.0, but your Gemfile requires did_you_mean 1.0.3. Prepending `bundle exec` to your command may solve this.
  /home/travis/.rvm/gems/ruby-2.3.4/gems/bundler-1.15.4/lib/bundler/runtime.rb:317:in `check_for_activated_spec!'
  /home/travis/.rvm/gems/ruby-2.3.4/gems/bundler-1.15.4/lib/bundler/runtime.rb:32:in `block in setup'
  /home/travis/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/forwardable.rb:204:in `each'
  /home/travis/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/forwardable.rb:204:in `each'
  /home/travis/.rvm/gems/ruby-2.3.4/gems/bundler-1.15.4/lib/bundler/runtime.rb:27:in `map'
  /home/travis/.rvm/gems/ruby-2.3.4/gems/bundler-1.15.4/lib/bundler/runtime.rb:27:in `setup'
  /home/travis/.rvm/gems/ruby-2.3.4/gems/bundler-1.15.4/lib/bundler.rb:101:in `setup'
  /home/travis/.rvm/gems/ruby-2.3.4/gems/bundler-1.15.4/lib/bundler/setup.rb:19:in `<top (required)>'
  /home/travis/.rvm/rubies/ruby-2.3.4/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
  /home/travis/.rvm/rubies/ruby-2.3.4/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
```

What I think happens is this:

[...]

PR #493, opened 2017-09-07
Move yard-junk out of :test group

mvz:
## Summary

This is a workaround for #492.

## Details

By not installing `yard-junk` on Travis, we avoid installing a newer version of `did_you_mean` in the vendor directory, and hence avoid an activation error.

## Types of changes


mvz:
Builds still fail on OSX, but that is unrelated.

PR #497, opened 2017-09-08
Fix final OSX build failures

mvz:
## Summary

This fixes/skips the cucumber scenarios failing on OSX.

## Details

The following bits need fixing:

* [x] $HOME is not in `/home` on OSX, so features that assume that need updating
* [x] Readline doesn't seem to be available on OSX for CRuby versions 1.9.3 and 2.0.0.
* ~Anything else that may turn up :-)~

~I have trimmed `.travis.yml` to be nice to Travis while this is still WIP. I'll remove that commit at the end.~

## Motivation and Context

Making master green ...

## Types of changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,10 @@ matrix:
     - rvm: ruby-head
     - rvm: jruby
     - rvm: jruby-9.1.13.0
+    - rvm: 1.9.3
+      os: osx
+    - rvm: 2.0.0
+      os: osx
   fast_finish: true
 branches:
   only:
--- a/bin/console
+++ b/bin/console
@@ -2,15 +2,6 @@
 
 $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
 
-require 'pry'
+require 'aruba/console'
 
-require 'aruba/api'
-
-module Aruba
-  class MyConsole
-    include Aruba::Api
-  end
-end
-
-include Aruba
-Pry.start MyConsole.new
+Aruba::Console.new.start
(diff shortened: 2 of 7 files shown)


mvz:
Travis seems to be thoroughly confused ...

mvz:
Travis is experiencing a major outage of OSX builds right now. I guess we'll have to wait ...

maxmeyer:
Why did you leave require readline as hook? Are there any places left where it makes sense? does this hook cause failures on Mac OS X?

mvz:
The scenarios marked with the `@requires-readline` hook need readline to work (because pry needs it). Apparently readline is not installed properly for older Rubies in Travis' OSX images (I don't know about real world OSX installations).

mvz:
(An alternative is to drop pry as a build dependency and just test with irb. That would need an update to `bin/console` too.)

maxmeyer:
@mvz I think it make sense to use a minimal approach an use IRB instead of pry although it doesn't have all the convenient features.

maxmeyer:
@mvz There are still failures on Mac OS X with 1.9.3 and 2.0.0. Do you own a mac to have a look for those?

mvz:
No, I don't. This is unfortunately very hard to debug.

maxmeyer:
Ok. I think we should merge it then. Are you online in slack? I don't want to litter this PR with conversation.

mvz:
Yes, I'm online.

mvz:
The OSX builds are green now.

PR #499, opened 2017-09-15
Make "have_output_size" accept a process, not a string

maxmeyer:
## Summary

Make the `have_output_size`-matcher work on a process output

## Details

This PR makes the matcher accept a process object and work on its output.

## Motivation and Context

In "Aruba" most API calls are about processes and checking their behaviour. The `have_output_size` matcher works on a string. 

## How Has This Been Tested?

New tests were added.

## Types of changes

- [ ] Update documentation

## Checklist:

### DIFF
--- a/lib/aruba/matchers/command/have_output_size.rb
+++ b/lib/aruba/matchers/command/have_output_size.rb
@@ -18,10 +18,15 @@
 #     end
 RSpec::Matchers.define :have_output_size do |expected|
   match do |actual|
-    next false unless actual.respond_to? :size
+    @old_actual = actual
 
-    @actual = actual.size
-    values_match? expected, @actual
+    next false unless @old_actual.respond_to? :output
+
+    @old_actual.stop
+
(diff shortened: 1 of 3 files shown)


maxmeyer:
@mvz As a user: What would you prefer size of  "sanitized" output or raw output?

maxmeyer:
If we decide to sanitize text I would add some more scenarios to make this clear in our documentation.

lib/aruba/matchers/command/have_output_size.rb
```diff
@@ -18,10 +18,15 @@
 #     end
 RSpec::Matchers.define :have_output_size do |expected|
   match do |actual|
-    next false unless actual.respond_to? :size
+    @old_actual = actual
 
-    @actual = actual.size
-    values_match? expected, @actual
+    next false unless @old_actual.respond_to? :output
```
mvz: I would prefer a hard crash in such cases.

PR #515, opened 2017-11-12
Mark scenarios requiring external commands

### LINKED ISSUE #513: Really skip zsh/python/etc tests
## Summary

Some cucumber scenarios requiring external commands like zsh, python, etc., are not marked as such.

## Expected Behavior

If the user does not have zsh installed, all scenarios requiring zsh should be skipped.

## Current Behavior

If the user does not have zsh installed, all scenarios requiring zsh are not skipped.

## Context & Motivation

Remark in https://github.com/cucumber/aruba/pull/512.

mvz:
## Summary

Fixes #513 

## Details

To avoid cucumber failures, marks scenarios requiring zsh, bash, python and ruby as such.

Also, avoids changing PATH when running ad-hoc scripts.

## Motivation and Context

See #513.

## Types of changes

### DIFF
--- a/features/03_testing_frameworks/cucumber/steps/command/run_commands_which_require_a_shell.feature
+++ b/features/03_testing_frameworks/cucumber/steps/command/run_commands_which_require_a_shell.feature
@@ -4,13 +4,15 @@ Feature: Running shell commands
   - `When I run the following script:`
 
   Or you can run shell commands with:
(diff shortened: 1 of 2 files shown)


olleolleolle:
I have a general question on syntax: ` spec.required_ruby_version = '>= 1.9.3'` is in the gemspec.

Could we begin using this Hash syntax? `{ foo: :bar, baz: 123 }`

PR #517, opened 2017-11-22
Prevent duplicate output

### LINKED ISSUE #374: Prevent duplicating stderr/stdout for `run_simple` commands

mvz:
## Summary

Ensure command output is only announced once.

## Details

During a run, `Command#stop` and `Command#terminate` are called multiple times, among other reasons to ensure all commands are actually terminated at the end of each spec or scenario.

If both general announcers and the announce-on-error functionality is used at the same time, duplicate output may still occur.

This includes changes from #399.

## Motivation and Context

This fixes #374.

## How Has This Been Tested?

Specs were added by @maxmeyer and myself to ensure stop events are only broadcast once on the event bus. The behavior was confirmed by running some of Aruba's features by hand and examining the entire output produced.

## Types of changes

## Checklist:

### DIFF
--- a/spec/aruba/command_spec.rb
+++ b/spec/aruba/command_spec.rb
@@ -24,67 +24,62 @@ RSpec.describe Aruba::Command do
   let(:stop_signal) { nil }
   let(:startup_wait_time) { 1 }
 
-  context '#started' do
-    before :each do
+  describe '#start' do
+    before do
       allow(event_bus).to receive(:notify).with(Events::CommandStarted)
     end
 
-    before :each do
-      command.start
-    end
-
     it 'leaves the command in the started state' do
+      command.start
       expect(command).to be_started
     end
   end
 
-  context '#stopped' do
-    before :each do
+  describe '#stop' do
+    before do
       allow(event_bus).to receive(:notify).with(Events::CommandStarted)
       allow(event_bus).to receive(:notify).with(Events::CommandStopped)
-    end
-
-    before :each do
       command.start
-      command.stop
     end
 
     it 'leaves the command in the stopped state' do
+      command.stop
       expect(command).to be_stopped
     end
 
     it 'notifies the event bus only once per run' do
+      command.stop
       command.stop
       expect(event_bus).to have_received(:notify).with(Events::CommandStopped).once
     end
 
     it 'prevents #terminate from notifying the event bus' do
+      command.stop
       command.terminate
       expect(event_bus).to have_received(:notify).with(Events::CommandStopped).once
     end
   end
 
-  context '#terminate' do
-    before :each do
+  describe '#terminate' do
+    before do
       allow(event_bus).to receive(:notify).with(Events::CommandStarted)
       allow(event_bus).to receive(:notify).with(Events::CommandStopped)
-    end
-
-    before :each do
       command.start
-      command.terminate
     end
 
     it 'leaves the command in the stopped state' do
+      command.terminate
       expect(command).to be_stopped
     end
 
     it 'notifies the event bus only once per run' do
       command.terminate
+      command.terminate
       expect(event_bus).to have_received(:notify).with(Events::CommandStopped).once
     end
 
     it 'prevents #stop from notifying the event bus' do
+      command.terminate
       command.stop
       expect(event_bus).to have_received(:notify).with(Events::CommandStopped).once
     end
--- a/spec/aruba/command_spec.rb
+++ b/spec/aruba/command_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 RSpec.describe Aruba::Command do
-  subject do
+  let(:command) do
     described_class.new(
       'true',
       event_bus: event_bus,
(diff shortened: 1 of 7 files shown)


spec/aruba/command_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  let(:startup_wait_time) { 1 }
+
+  context '#started' do
+    before :each do
+      allow(event_bus).to receive(:notify).with(Events::CommandStarted)
+    end
+
+    before :each do
+      command.start
+    end
```
olleolleolle: Communication: I think this is the "Act" step of Arrange-Act-Assert, and that it's not a `before`-test setup.

Suggestion: Move `command.start` into the `it`.

spec/aruba/command_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  let(:working_directory) { expand_path('.') }
+  let(:environment) { ENV.to_hash }
+  let(:main_class) { nil }
+  let(:stop_signal) { nil }
+  let(:startup_wait_time) { 1 }
+
+  context '#started' do
+    before :each do
+      allow(event_bus).to receive(:notify).with(Events::CommandStarted)
+    end
```
olleolleolle: `before :each` => `before` (this is the default; and it has been renamed in newer versions of RSpec; [`:each` is now `:example`](https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/before-and-after-hooks))

mvz:
@olleolleolle I made the changes you suggested, thanks.

olleolleolle:
@mvz It's a pleasure to learn so much from the Aruba work you do here.

maxmeyer:
@mvz Personally I prefer this type of specs, which are more "concise" as they use the Rspec syntax to define readable tests: http://www.betterspecs.org/.

mvz:
@maxmeyer I have two problems with those type of tests:
* Using `subject` hides the nature of the object under test. Giving it an expressive name allows you to interpret the expectation itself more easily.
* Using an anonymous `it` block makes the intent of the test less clear: Yes, the method has the expected result, but why do we want it to have that result? Also, some desired outcomes cannot be expressed in RSpec syntax (e.g.,`it 'prevents #stop from notifying the event bus').

olleolleolle:
@mvz In order to have the cake and eat it, explicitly naming the `subject` allows 


```ruby
let(:command) do
  described_class.new(
    'true',
    event_bus: event_bus,
    exit_timeout: exit_timeout,
    io_wait_timeout: io_wait_timeout,
    working_directory: working_directory,
    environment: environment,
    main_class: main_class,
    stop_signal: stop_signal,
    startup_wait_time: startup_wait_time
  )
end 
```

to be

```ruby
subject(:command) do
  described_class.new(
    'true',
    event_bus: event_bus,
    exit_timeout: exit_timeout,
    io_wait_timeout: io_wait_timeout,
    working_directory: working_directory,
    environment: environment,
    main_class: main_class,
    stop_signal: stop_signal,
    startup_wait_time: startup_wait_time
  )
end 
```

and, taking it a step further, passing the constructor parameters we're not manipulating in the test as primitive values would reduce "setup".

maxmeyer:
@mvz Sorry, I was not very clear about that. I was a bit in a rush and didn't see, that that use `is_expected`. I don't like that style either, but try to avoid this.

I try to avoid using strings in the `it`-method.

~~~ruby
RSpec.describe MyClass do
  it 'some description' do
     expect(true).to_be true
  end
end
~~~

Instead I try to use something like that:

~~~ruby
RSpec.describe MyClass do
  subject(:my_instance) { described_class.new }

  context 'when action was run' do
    # Set state - if required
    before :example do
      my_instance.my_action
    end

    # Assert
     it { expect(my_instance.result).to_be true }
  end
end
~~~

I'm open to change the style to something which suits our needs best. But besides that, what I really want to see for our test suite: Use the same style we can commit to everywhere and don't mix styles. Example for the style current used at most places I had a look at: https://github.com/cucumber/aruba/blob/master/spec/aruba/api_spec.rb#L8-L18 vs.

mvz:
@maxmeyer agreed on having a consistent style. I'll rewrite the examples and see how that goes.

olleolleolle:
@maxmeyer Hi!

[rspec-its](https://github.com/rspec/rspec-its) is now a separate RubyGem.

Myron Marston wrote this, before its was removed: [Explanation for why `its` will be removed from rspec-3
](https://gist.github.com/myronmarston/4503509)

maxmeyer:
@olleolleolle Thanks for clarifying that. Is there any place where we use this kind of style? I searched the repository and didn't find a place where we do this. Did I miss something?

First, I thought http://www.betterspecs.org/ recommends this style, but I can't find anything about `rspec-its` there either. It's nothing I used before. And reading Myron's writeup makes me thing, this is not a good way to go.

My style of writing RSpec specs was influenced by a talk. I think, it was this one: https://www.youtube.com/watch?v=d2gL6CYaYwQ. Though I don't use the `is_expected`-syntax for the reasons @mvz wrote down. 

The style recommended by this talk helps me to 
* concentrate on the way my code behaves in different contexts - using `#context "when something is given"`
* to DRY up my tests - though is goes "against" readability

But as you both are more into testing than me - my daily job is to keep things running as a sysadmin - I think you've got more experience than me and I'm happy to learn new things. :smile:

spec/aruba/command_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'spec_helper'
+
+RSpec.describe Aruba::Command do
+  let(:command) do
```
maxmeyer: As discussed: I prefer to use `subject(:command)` here.

spec/aruba/command_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  let(:stop_signal) { nil }
+  let(:startup_wait_time) { 1 }
+
+  describe '#start' do
+    before do
+      allow(event_bus).to receive(:notify).with(Events::CommandStarted)
+    end
+
+    it 'leaves the command in the started state' do
+      command.start
```
maxmeyer: I would place this in a separate `#before`-block`.

olleolleolle:
@maxmeyer Now that we're trading video links, here's Searls, talking (very quickly) about testing:

- https://www.youtube.com/watch?v=aiDikM9g4x8
- https://www.youtube.com/watch?v=VD51AkG8EZw

mvz:
@olleolleolle I updated the specs added one little fix for the InProcess runner. Can you take one final look at this?

olleolleolle:
Great! Covering those bits is super good.

Issue #490, opened 2017-09-04
Using Aruba in directories with spaces


odlp:
## Summary

Aruba (`0.14.2`) isn't able to spawn processes within folders which have spaces in the name.

## Expected Behavior

```sh
$ pwd
/Users/oli/workspace/aruba-spaces-bug/regrettable folder with spaces

$ bundle exec cucumber -b

Feature: Example

  Scenario: Running in a directory with spaces # features/example.feature:2
    When I successfully run `cli`              # aruba-0.14.2/lib/aruba/cucumber/command.rb:27
    Then the output should contain:            # aruba-0.14.2/lib/aruba/cucumber/command.rb:205
      """
      foo
      """

1 scenario (1 passed)
2 steps (2 passed)
0m0.113s
```

## Current Behavior

```sh
$ pwd
/Users/oli/workspace/aruba-spaces-bug/regrettable folder with spaces

$ bundle exec cucumber -b

Feature: Example

  Scenario: Running in a directory with spaces # features/example.feature:2
    When I successfully run `cli`              # aruba-0.14.2/lib/aruba/cucumber/command.rb:27
      undefined local variable or method `cmd' for #<Aruba::Processes::SpawnProcess:0x007f95102b2328>
      Did you mean?  @cmd (NameError)
      /Users/oli/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/aruba-0.14.2/lib/aruba/processes/spawn_process.rb:86:in `rescue in start'
      /Users/oli/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/aruba-0.14.2/lib/aruba/processes/spawn_process.rb:80:in `start'
      /Users/oli/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/aruba-0.14.2/lib/aruba/command.rb:67:in `start'
      /Users/oli/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/aruba-0.14.2/lib/aruba/api/command.rb:213:in `run'
      /Users/oli/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/aruba-0.14.2/lib/aruba/api/command.rb:271:in `run_simple'
      /Users/oli/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/aruba-0.14.2/lib/aruba/cucumber/command.rb:29:in `block in <top (required)>'
[...]

PR #520, opened 2017-11-28
Fix command spawning when spaces occur in the path

mvz:
## Summary

There was a bug in how `ChildProcess.build` was called. This PR fixes this.

## Details

Calling `#to_a` was called on UnixCommandString representing the executable, splitting the name unnecessarily.

## Motivation and Context

See #490.

## How Has This Been Tested?

A spec was added for this. This still needs to be tested on Windows.

## Types of changes

## Checklist:


mvz:
This may need a slightly diffferent approach to not break on Windows.

mvz:
This is ready for review.

mvz:
@olleolleolle ok to merge?

lib/aruba/platforms/windows_command_string.rb
```diff
@@ -11,13 +11,18 @@ module Platforms
     # @private
     class WindowsCommandString < SimpleDelegator
       def initialize(cmd)
-        __setobj__ format('%s /c "%s"', Aruba.platform.which('cmd.exe'), cmd)
+        __setobj__ cmd
```
olleolleolle: This is the default right? No explicit need to implement the constructor?

mvz: You may be right. I'll take a look.

mvz: Yes, it is no longer needed. I removed it.

PR #529, opened 2018-01-19
fix @no-clobber breaking process management (closes 

doudou:
## Summary

When a scenario has @no-clobber, Aruba would not call Setup#call *at all*, which causes the
process monitor to not be listening to process events.

## Details

Changed Setup#call to accept a 'clobber' argument and set it according to whether there is @no-clobber or ~@no-clobber.

## Motivation and Context

Closes #527

## How Has This Been Tested?

**THERE ARE NO TESTS YET** I'm just not sure where it is appropriate to test for this. It's weird to do this in a cucumber feature, as it's testing a regression and not a feature, but then we should check that @no-clobber and process monitoring work (which can only be done in cucumber). It's going to look weird in the cucumber.pro site, no ?

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactoring (cleanup of codebase withouth changing any existing functionality)
- [ ] Update documentation

## Checklist:

- [ ] I've added tests for my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #533, opened 2018-02-02
FIx syntax to work with old rubies

mvz:
## Summary

The syntax in #529 and #490 is not compatible with Ruby 1.8.3. This PR fixes that.

## Types of changes


mvz:
I added another fix.

PR #535, opened 2018-02-02
Fix @no-clobber behavior

### LINKED ISSUE #527: @no-clobber may break process monitoring
I have a cucumber feature file where all the scenarios have @no-clobber, using aruba 0.14.3. This is because the very first feature bootstraps an install and is very expensive. The followup features only use it.

In this file, the steps that involve process management (I should stop, matching on output, ...) do not work (the output is always empty, waiting for process stop always complains that there are no processes started).

I've been putting some strategic `puts`, and it seems that the process monitor does not get notified of new commands being started. I think that I tracked this down to Setup#call not being called (it's called only in `Before("~@no-clobber")`, but Setup#events is what links the process monitor to the events.

mvz:
## Summary

Ensure setup is run even if files are not clobbered.

## Details

See #529.

## Motivation and Context

Fixes #527.

## How Has This Been Tested?

Scenarios were updated to demonstrate that setup is also run when `@no-clobber` is set, but files are not clobbered.

## Types of changes

- [x] Update documentation

## Checklist:

### DIFF
--- a/lib/aruba/api/core.rb
+++ b/lib/aruba/api/core.rb
@@ -25,8 +25,8 @@ module Aruba
       # This will only clean up aruba's working directory to remove all
       # artifacts of your tests. This does NOT clean up the current working
       # directory.
-      def setup_aruba
-        Aruba::Setup.new(aruba).call
+      def setup_aruba(clobber: true)
+        Aruba::Setup.new(aruba).call(clobber: clobber)
 
         self
       end
--- a/lib/aruba/api/core.rb
+++ b/lib/aruba/api/core.rb
@@ -25,8 +25,8 @@ module Aruba
       # This will only clean up aruba's working directory to remove all
       # artifacts of your tests. This does NOT clean up the current working
       # directory.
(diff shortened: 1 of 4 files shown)


lib/aruba/api/core.rb
```diff
@@ -25,8 +25,8 @@ def aruba
       # This will only clean up aruba's working directory to remove all
       # artifacts of your tests. This does NOT clean up the current working
       # directory.
-      def setup_aruba
-        Aruba::Setup.new(aruba).call
+      def setup_aruba(clobber: true)
+        Aruba::Setup.new(aruba).call(clobber: clobber)
```
olleolleolle: This syntax is the incompatible one.

mvz: Yes, but it should be fine on master since we dropped support for 1.8.

mvz: Ah bummer, I forgot this won't work with 1.9 either 😞

PR #540, opened 2018-02-11
Restore old expand path behavior

### LINKED ISSUE #478: Absolute directory paths and existence check
First thanks for aruba :tada: 

## Summary

I upgraded aruba in simplecov from 0.7 to 0.14.2. One spec started failing, basically it configures a different directory for simplecovs output as:

      SimpleCov.start do
        coverage_dir '/tmp/test/simplecov'
      end

The test asserting that the directory exists started failing, and I fear that it also won't find the files I'm checking for.

It is imo because the method:

```ruby
      def directory?(file)
        Aruba.platform.directory? expand_path(file)
      end
```

calls out to `expand_path` which gets the path wrong:

```
[1] pry(#<Object>)> file
=> "/tmp/test/simplecov"
[2] pry(#<Object>)> expand_path(file)
=> "/home/tobi/github/simplecov/tmp/aruba/project/tmp/test/simplecov"
```
So `/tmp/test/simplecov` exists (and we created it using `File.mkdir_p`) but the path aruba thinks that it should be does not exist, hence the test fails ([this is the test](https://github.com/colszowka/simplecov/blob/master/features/config_coverage_dir.feature#L22-L33))

## Expected Behavior

I'd expect aruba to find directories/files created by my application in the tests. 

## Current Behavior

Aruba looks in its working directory not the actual absolute path.

## Possible Solution

One could check if the path is absolute and then just not expand it and everything should work fine.

[scenario failing when updating aruba](https://github.com/colszowka/simplecov/blob/master/features/config_coverage_dir.feature#L22-L33)

Basically just running `directory?('/some/absolute/path')` seems to fail for me.

## Context & Motivation

People might wanna put their simplecov coverage reports in an absolute path to gather them or whatever. I wanna make sure they can do that.

## Your Environment

* Version used: 0.14.2
* Operating System and version:
  * Linux Mint 18.2
  * `uname -a`: Linux speedy 4.10.0-32-generic #36~16.04.1-Ubuntu SMP Wed Aug 9 09:19:02 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Link to your project: https://github.com/colszowka/simplecov
* Ruby: 2.4

mvz:
## Summary

Allow absolute paths in `#expand_path`, but complain about it by default.

## Details

Makes `#expand_path` correctly handle absolute paths (like in #539), but in addition emit a warning when an absolute path is passed. Adds a configuration option `allow_absolute_paths` that can be set to `true` to silence this warning.

## Motivation and Context

Fixes #478.

## How Has This Been Tested?

Specs were added for the case of handing absolute paths, for emission of a warning, and for silencing that warning.

## Types of changes

## Checklist:


spec/spec_helper.rb
```diff
@@ -13,6 +13,3 @@
 # Loading support files
 Dir.glob(::File.expand_path('../support/*.rb', __FILE__)).each { |f| require_relative f }
 Dir.glob(::File.expand_path('../support/**/*.rb', __FILE__)).each { |f| require_relative f }
-
-# Avoid writing "describe LocalPac::MyClass do [..]" but "describe MyClass do [..]"
-include Aruba
```
olleolleolle: Thank you for making _this change_.

mvz:
Thanks, @olleolleolle. I fixed most of the odd wording.

matthutchinson:
Great work! thanks for merging 👍

matthutchinson:
Will this be pushed out in a 0.14 point release (soon) or as part of 1.0?

mvz:
@matthutchinson This partiular PR will be part of 1.0, but #539 implements just the fix to `expand_path` and will be part of the next 0.14 point release.

PR #543, opened 2018-03-01
Remove two less-useful scenarios

mvz:
## Summary

Remove some scenarios that are less useful.

## Details and Motivation

* Remove the scenario that basically runs `gem install aruba`. This is not really a feature of Aruba itself, installation is already explained in the README, and this scenario takes quite some time on CI.
* Remove the scenario explaining how to write Markdown in Cucumber feature files. This was relevant for the Aruba reporting option, but this has been removed for 1.0.0 so now it makes no sense anymore.

## Types of changes

- Update documentation


olleolleolle:
Nice! Pruning the test suite is important work.

PR #553, opened 2018-03-16
Loosen dependency on childprocess

mvz:
## Summary

Like it says on the tin.

## Details

Allows all versions of childprocess that we can reasonable expect to work with aruba according to the [childprocess changelog](https://github.com/enkessler/childprocess/blob/master/CHANGELOG.md)

## Motivation and Context

Allows latest childprocess to be used with Aruba (like already is the case for 0.14.x).

## How Has This Been Tested?

We're currently lacking tests for checking if the lowest required versions for dependencies still work. Other than that: Travis.

## Types of changes

- [ ] Update documentation

## Checklist:

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.homepage    = 'http://github.com/cucumber/aruba'
 
   spec.add_runtime_dependency 'cucumber', '~> 2.4', '>= 2.4.0'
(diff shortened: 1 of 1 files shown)


enkessler:
FYI: `childprocess` is semantically versioned (or at least has been since I've been maintaining it), so you should be safe all the way until `1.0.0`.

mvz:
@enkessler good to know. I'll take that into account in the next dependency update.

PR #571, opened 2018-08-27
Mark setting of root directory as deprecated

mvz:
## Summary

Setting root_directory will no longer be supported in 1.0.0. Mark it as deprecated in the cucumber scenario.

**This was originally the main pull request for #598. Original description follows**

## Summary

Update the `still` development branch so the change to `master` is no more than a removal of deprecations.

## Details

**This is a work in progress**. This is a combination of backporting bug fixes, moving deprecated code, moving other files. The end goal is that merging `still` into `master` will only lead to conflicts in deleted files that contain the code deprecated in the `0.14.x` line and removed in `1.0.0`.

## Motivation and Context

A lot of projects that use Aruba are not even on `0.14.x` yet. We should be able to tell them that they should do that and fix all deprecations, and then they should be able to safely move to upcoming `1.0.0`. Since there are still reasons why projects are stuck on pre-0.14 versions, these reasons should be taken away first. This means Aruba development is stuk on 0.14 for a while longer as well. To avoid duplicate effort, `still` and `master` need to be made as similar as possible.

We may have to revert some breaking behavioral changes between still and master eventually, as well as some deprecations.

## How Has This Been Tested?

CI. I'm hoping to add AppVeyor to the still branch as well.

## Types of changes

- [x] Refactoring (cleanup of codebase without changing any existing functionality)
- [x] Update documentation

## Checklist:

### DIFF
--- a/features/02_configure_aruba/command_runtime_environment.feature
+++ b/features/02_configure_aruba/command_runtime_environment.feature
@@ -23,7 +23,7 @@ Feature: Define default process environment
     ENV['LONG_LONG_VARIABLE'] = 'y'
 
     Aruba.configure do |config|
-      config.command_runtime_environment = { 'LONG_LONG_VARIABLE' => 'x' } 
+      config.command_runtime_environment = { 'LONG_LONG_VARIABLE' => 'x' }
     end
 
     RSpec.describe 'Environment command', :type => :aruba do
@@ -44,7 +44,7 @@ Feature: Define default process environment
     ENV['LONG_LONG_VARIABLE'] = 'y'
 
     Aruba.configure do |config|
-      config.command_runtime_environment = { 'LONG_LONG_VARIABLE' => 'x' } 
+      config.command_runtime_environment = { 'LONG_LONG_VARIABLE' => 'x' }
     end
 
     RSpec.describe 'Environment command', :type => :aruba do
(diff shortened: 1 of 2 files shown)


xtrasimplicity:
This is a great idea, @mvz! Shout out if you'd like a hand and I'll see what I can do. :)

mvz:
I'm going to move parts of this to separate pull requests, to be coordinated by #598.

luke-hill:
What's the benefit of these 2 branches and why don't we just have 1 master branch?

Is it because we're supporting a 1.0 and 0.9 based development, if so could we not rename the 0.9 branch as something like 0_9_stable (Akin to rails). And have the 1.0 stuff on master

mvz:
Yeah, 1.0 stuff is on master and 0.14.x stuff is on still. The original plan (from before my time) was to have a 1.0 version a lot more quickly, but a lot of projects that use Aruba are still not even on 0.14.x.

Sooo ... I'm first going to focus on helping them upgrade and fixing stuff in 0.14.x for that.

But since I don't want to keep forward-porting bug fixes between these two branches, I want to get to a point where I can at least merge still into master regularly.

Renaming still to something else would be fine.

mvz:
This is now only a documentation change. Merging.

PR #577, opened 2018-12-28
Update build matrix

mvz:
## Summary

Update build matrix with recent Rubies

## Details

Update Travis CI configuration to run tests on Rubies of the 2.4, 2.5 and 2.6 series.

## Motivation and Context

Ruby 2.6 has been released and we should test Aruba on it.


mvz:
Yeah, we should update the jruby set as well.

PR #578, opened 2018-12-28
Test with Ruby 2.6

mvz:
## Summary

Update build matrix with Ruby 2.6.

## Details

Update Travis CI configuration to run tests on Ruby 2.6.0 series.

## Motivation and Context

Ruby 2.6 has been released and we should test Aruba on it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,11 @@ group :development, :test do
   end
 
   # Run development and test tasks
-  gem 'rake', '~> 12.0'
+  if RUBY_VERSION >= '2.0.0'
+    gem 'rake', '~> 12.3'
+  else
+    gem 'rake', '~> 12.2.0'
+  end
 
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -44,7 +44,11 @@ group :development, :test do
   end
 
   # Run development and test tasks
-  gem 'rake', '~> 12.0'
+  if RUBY_VERSION >= '2.0.0'
+    gem 'rake', '~> 12.3'
+  else
+    gem 'rake', '~> 12.2.0'
+  end
```
olleolleolle: Sometimes, a fiddlier way of expressing the same thing avoids warnings output from Bundler. Like a ternary conditional operator, etc.

olleolleolle: I _have had_ warnings from Bundler telling me not to repeat myself. Perhaps they're _smarter_ now, perhaps that was an edge-case.

PR #581, opened 2019-01-07
Update dependencies

mvz:
## Summary

Update dependencies

## Details

Update dependency on FFI and several development dependencies.

## Motivation and Context

FFI 1.10 is out and Aruba should allow users to use it. For development dependencies, simply ensures keeping up to date.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,7 +82,7 @@ group :development, :test do
     gem 'contracts', '~> 0.16.0'
   end
 
-  if RUBY_VERSION >= '1.9.3'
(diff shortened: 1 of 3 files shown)


olleolleolle:
Perhaps you can use `gem install bundler:'<2'`?

mvz:
> Perhaps you can use `gem install bundler:'<2'`?

I made a little tighter version requirement, but still allowing for patch releases of Bundler 1.17.

olleolleolle:
Thanks for polishing this!

PR #595, opened 2019-02-14
Actionable code coverage

grosser:
Atm every runs finishes with `Coverage report generated for rspec to coverage. 1187 / 2102 LOC (56.47%) covered.` which is kinda useless and wastes time on every test run.

I'd suggest switching to [single_cov](https://github.com/grosser/single_cov) which is:
 - much faster
 - supports branch coverage
 - returns actionable feedback

```
lib/aruba/processes/in_process.rb new uncovered lines introduced (11 current vs 0 configured)
Lines missing coverage:
lib/aruba/processes/in_process.rb:20
lib/aruba/processes/in_process.rb:36
lib/aruba/processes/in_process.rb:56:9-56:72
lib/aruba/processes/in_process.rb:69:11-69:22
lib/aruba/processes/in_process.rb:81
lib/aruba/processes/in_process.rb:99
lib/aruba/processes/in_process.rb:104
lib/aruba/processes/in_process.rb:106
lib/aruba/processes/in_process.rb:111
lib/aruba/processes/in_process.rb:118
lib/aruba/processes/in_process.rb:122
```


olleolleolle:
Here's the _pretty_ output:

```
lib/aruba/processes/in_process.rb new uncovered lines introduced (9 current vs 0 configured)
Lines missing coverage:
lib/aruba/processes/in_process.rb:36
lib/aruba/processes/in_process.rb:56:9-56:72
lib/aruba/processes/in_process.rb:69:11-69:21
lib/aruba/processes/in_process.rb:81
lib/aruba/processes/in_process.rb:99
lib/aruba/processes/in_process.rb:104
lib/aruba/processes/in_process.rb:106
lib/aruba/processes/in_process.rb:111
lib/aruba/processes/in_process.rb:118
```

grosser:
not sure what you are trying to say with this comment, I was hoping for a "this looks good, let's do this for everything" or "nah, we'll keep it like it is"

olleolleolle:
@mvz Do you agree with that?

mvz:
I'll play around with this tool a bit in the coming weeks to see how it compares to simplecov.

A problem for Aruba is that most of the code is covered by cucumber scenarios. I'd like to know how single_cov deals with that. I'd also like to know how to deal with test coverage that is not one-to-one in terms of files.

Are there any benchmarks demonstrating the simplecov overhead? I thought its overhead was supposed to be very small.

mvz:
Final question: What would be needed to make tests green for this pull request?

grosser:
single_cov does not support cucumber, so that won't work :)
it supports not having a 1-1 mapping, but then it regresses to being as bad
as simplecov since it will report weird results when running
individual results
so should only use if have 1-1 unit test coverage is a goal, can always do
more with cucumber etc, but that won't be counted

PR #600, opened 2019-03-13
Add support for have_output* with no arguments

peagha:
## Summary

Allow have_output* matchers to work with no arguments

## Details

When using the have_output* matchers they will fail to match if no argument is provided:

```
expect(last_command_stopped).not_to have_output_in_stdout
```
The documentation has examples like that:
https://github.com/cucumber/aruba/blob/8d02d98f7acc6b985f42917d1f025f1d060a8022/lib/aruba/matchers/command/have_output_on_stderr.rb#L13-L15
https://github.com/cucumber/aruba/blob/8d02d98f7acc6b985f42917d1f025f1d060a8022/lib/aruba/matchers/command/have_output.rb#L15-L17

## How Has This Been Tested?

RSpec examples were added to cover the changes

## Types of changes

- [ ] Update documentation

## Checklist:


peagha:
I've noticed the failure message is not working properly; I'll fix that and push again

mvz:
Hi @peagha, you seem to have removed the 'motivation and context' section from the pull request template. Can you tell me a little bit more about where you encountered problems with the current behavior?

Also, which version of Aruba are you using?

peagha:
@mvz thanks for your answer!

The main motivation is that I had a scenario where the command was failing (exit status != 0) and I wanted to have feedback on that. Older versions of Aruba used to print the whole output when using `assert_passing_with`, like that:
``` 
Exit status was 1 but expected it to be 0. Output:
      /Users/peagha/Code/me/forca/bin/forca:5:in `<main>': unhandled exception
```
In the latest version, when using `expect(last_command_stopped).to be_successfully_executed`, all I get is:
```
(...)
expected "forca" to be successfully executed (RSpec::Expectations::ExpectationNotMetError)
(...)
```

The output is useful to understand why the command wasn't successfully executed, therefore I added to my app:
```ruby
expect(last_command_stopped).not_to have_output_on_stderr
```

What would print the STDERR whenever something is wrong. But that doesn't work because `have_output_on_stderr` expects an argument. That motivated me to change the matcher to work without arguments 😄 

I'm using the lates version of Aruba (`0.14.8`)

peagha:
I've updated the failure messages code and they should be working now

PR #604, opened 2019-03-17
Do not replace entire environment inside cd block

mvz:
## Summary

Backport fix for `cd` behavior.

## Details

Test and fix the the setting of environment variables inside the block passed to `cd`. Backport of cbbefda22343ee8a858cc6e0ce85446b107b5609, which was part of #442.

## Motivation and Context

The `#cd` method would clear the entire environment within the block, instead of just updating it. This is clearly not the intended behavior, since even variables set with `set_environment_variable` are ignored. This also makes `#cd` a bad substitute for the deprecated `#in_current_directory` method.

## How Has This Been Tested?

Specs were added.

## Types of changes

## Checklist:

### DIFF
--- a/spec/aruba/api_spec.rb
+++ b/spec/aruba/api_spec.rb
@@ -113,7 +113,25 @@ describe Aruba::Api do
           expect(Dir.pwd).not_to eq full_path
         end
 
-        it 'does not touch non-directory environment the passed block' do
+        it 'sets directory environment in the passed block' do
+          @aruba.create_directory @directory_name
+          old_pwd = ENV['PWD']
+          full_path = File.expand_path(@directory_path)
+          @aruba.cd @directory_name do
+            expect(ENV['PWD']).to eq full_path
+            expect(ENV['OLDPWD']).to eq old_pwd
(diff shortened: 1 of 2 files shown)


spec/aruba/api_spec.rb
```diff
@@ -101,6 +101,26 @@
...
+        it 'runs the passed block in the given directory' do
+          @aruba.create_directory @directory_name
+          full_path = File.expand_path(@directory_path)
+          @aruba.cd @directory_name do
+            expect(Dir.pwd).to eq full_path
+          end
+          expect(Dir.pwd).not_to eq full_path
+        end
+
+        it 'does not touch non-directory environment the passed block' do
```
olleolleolle: The wording of this `it` seems like it's missing an `of` or a `the` somewhere. Can you try adding words?

mvz: I updated the wording and added some more specs.

olleolleolle: That is SUPER! The new wording is _crystal clear_.

PR #609, opened 2019-05-03
Release 1.0.0.alpha3

luke-hill:
Hi @mvz I cut a checkpoint roughly halfway between 1.0.alpha2 and the current master. And I've badged it as a 1.0.alpha.3

## Motivation and Context

Hopefully this gives us "some" of the newer code in a release so we can see how much of the newer stuff is working / not working.

Obviously the majority of stuff between 1.0.alpha2 and master is mainly test/dev tweaks, so I don't envision this will be "that" different too the 2nd alpha, but it includes a healthy amount of commits (Around 150-200).

Hopefully getting this cut and released will make the job of releasing a 4th alpha / stable 1.0 a bit easier, as we can see what stuff here works and what doesn't.

## Types of changes

- [ ] Update documentation

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project will adhere to [Semantic Versioning][1] once version 1.0.0 is relea
 
 This document is formatted according to the principles of [Keep A CHANGELOG][2].
 
-## [v1.0.0-alpha.3]
+## [1.0.0-alpha.3]
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


luke-hill:
Obviously you can just cut a release mark against this branch and release the gem, or you can merge this directly into master if you'd prefer and cut it that way (But it's deliberately several hundred commits behind)

mvz:
Why not release current master as alpha3?

luke-hill:
I figured that there's been like 500 commits and that there probably should have been an interim release. I guess you could if you felt as though the likelihood is there wouldn't be lots of more issues that would come out. Whatever you think is easiest.

mvz:
Ok, I'll try to get an alpha 3 out next weekend.

luke-hill:
If there's any code cutting / tasks you want me to do fire them my way, rails doesn't have much and I've got a few outstanding PRs needing review from matt in ruby, so not much else I can do here

mvz:
... and released it is :tada:!

PR #612, opened 2019-05-05
Fix JRuby before :command helper hook 

mvz:
## Summary

Add full integration test for JRuby helper and make it work.

## Details

The before :command hooks are run after the command's environment has already been set up, so instead of modifying Aruba's environment, of the actual environment, we need to modify the command's environment. This change adds a scenario for this feature, and updates the implementation
to make it work.

We also no longer require the helper by default, so requiring by the user will have the desired effect.

## Motivation and Context

An attempt was made to fix this feature in #610, but it lacked an integration test so it failed.

## How Has This Been Tested?

A scenario was added and the specs were updated.

## Types of changes

## Checklist:

### DIFF
--- a/features/08_other/improve_performance_if_using_jruby.feature
+++ b/features/08_other/improve_performance_if_using_jruby.feature
@@ -4,7 +4,7 @@ Feature: Support for JRuby
   be accomplished by adding
 
   ```ruby
-  require 'aruba/jruby'
+  require 'aruba/config/jruby'
   ```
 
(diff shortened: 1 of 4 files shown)


olleolleolle:
Oh, 1.8.7. Ouch! 

```
SyntaxError:
  /home/travis/build/cucumber/aruba/spec/aruba/jruby_spec.rb:9: syntax error, unexpected ':', expecting ')'
  ...es::BasicProcess, environment: command_environment)
```

PS: This work is super. Thank you for taking the time to build the JRuby support.

PR #628, opened 2019-05-20
TimeoutError is deprecated in favor of Timeout::Error

amatsuda:
Ruby stdlib Timeout's TimeoutError is deprecated in favor of Timeout::Error.

https://github.com/ruby/ruby/blob/34d7ec4e9b0a24b3ce89009927738f1dac38e6d3/lib/timeout.rb#L125-L130


amatsuda:
I mean, even in the oldest supported Ruby, 2.2.0 https://github.com/ruby/ruby/blob/v2_2_0/lib/timeout.rb#L128-L130

PR #631, opened 2019-05-23
Gemfile/Gemspec Tidy (For Master)

luke-hill:
## Summary

Tidy up Gemfile / Deps

## Details

Migrate everything that could be out of gemfile conditionals into gemspec fixed values.
NB: I've also updated a few minor version restrictions. And the *minimum ruby version* although this was always 2.3.3 due to a dependency requirement.

## Motivation and Context

Tidies things up.

## How Has This Been Tested?

CI

## Types of changes

- [ ] Update documentation

## Checklist:

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'yard-junk', '~> 0.0.7'
 
   spec.rubygems_version = '>= 1.6.1'
-  spec.required_ruby_version = '>= 2.3.3'
+  spec.required_ruby_version = '>= 2.3'
 
   spec.files = `git ls-files -z`.split("\x0").reject do |f|
     f.match(%r{^(test|spec|features)/})
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -21,23 +21,21 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'rspec-expectations', '~> 3.4'
   spec.add_runtime_dependency 'thor', '~> 0.19'
 
-  spec.add_development_dependency 'fuubar', '~> 2.2'
+  spec.add_development_dependency 'fuubar', '~> 2.3'
   spec.add_development_dependency 'json', '~> 2.1'
-  spec.add_development_dependency 'license_finder', '~> 5.0'
-  spec.add_development_dependency 'minitest', '~> 5.8'
-  spec.add_development_dependency 'pry', '~> 0.12.2'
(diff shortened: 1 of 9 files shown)


luke-hill:
I'll make one of these for `still` in a day or so.

luke-hill:
ping @mvz i've no idea if the CI here is stable or not? Could you please advise? Can this be merged?

mvz:
I think an adjustment to Travis' configuration is needed. Hold on.

aruba.gemspec
```diff
@@ -21,9 +19,21 @@ Gem::Specification.new do |spec|
...
 
-  spec.add_development_dependency 'bundler', ['>= 1.7.0', '< 3.0']
+  spec.add_development_dependency 'fuubar', '~> 2.3'
+  spec.add_development_dependency 'json', '~> 2.1'
+  spec.add_development_dependency 'license_finder', '~> 5.3'
+  spec.add_development_dependency 'minitest', '~> 5.10'
+  spec.add_development_dependency 'pry-doc', '~> 1.0.0'
+  spec.add_development_dependency 'rake', '~> 12.3'
+  spec.add_development_dependency 'rspec', '~> 3.6'
+  spec.add_development_dependency 'rubocop', '~> 0.69'
```
mvz: This should be `~> 0.69.0` because the set of cops can change with each release.

aruba.gemspec
```diff
@@ -21,9 +19,21 @@ Gem::Specification.new do |spec|
...
+  spec.add_development_dependency 'rubocop-performance', '~> 1.1'
+  spec.add_development_dependency 'simplecov', '~> 0.15'
+  spec.add_development_dependency 'travis-yaml', '~> 0.2'
+  spec.add_development_dependency 'yard-junk', '~> 0.0.7'
+
   spec.rubygems_version = '>= 1.6.1'
-  spec.required_ruby_version = '>= 2.3'
+  spec.required_ruby_version = '>= 2.3.3'
```
mvz: I'm not a fan of excluding micro versions in general. Can we avoid this? Which of our depedencies needs 2.3.3? Maybe we need to test on 2.3.0 in Travis as well?

mvz:
The build should be green now. The signs are good :slightly_smiling_face:.

aruba.gemspec
luke-hill: I agree with this, but I just copied the existing config, all this PR is aimed to be is a tidy up.

aruba.gemspec
luke-hill: One of the documentation ones in the development dependencies, if we reduce it down we should be ok

PR #634, opened 2019-05-30
Allow multiple arguments to xProcess#write

mvz:
## Summary

Allow passing more than one string to `InProcess#write` and `SpawnProcess#write`

## Details

Changes the arity of these two methods to match that of the base class' method. This makes these methods match the behavior of `IO#write`.

## Motivation and Context

This avoids the need to do a string append in `Api::Commands#type`, which in turn will help avoid issues when frozen strings are passed to that method. As a next step, all strings in Aruba can be made frozen by default.

## How Has This Been Tested?

Specs have been added

## Types of changes

- [ ] Update documentation

## Checklist:

### DIFF
--- a/lib/aruba/api/commands.rb
+++ b/lib/aruba/api/commands.rb
@@ -186,7 +186,7 @@ module Aruba
       def type(input)
         return close_input if input == "\u0004"
 
-        last_command_started.write(input << "\n")
+        last_command_started.write(input, "\n")
       end
 
       # Close stdin
--- a/lib/aruba/processes/basic_process.rb
+++ b/lib/aruba/processes/basic_process.rb
@@ -52,15 +52,15 @@ module Aruba
       end
 
       def stdin(*)
-        NotImplementedError
+        raise NotImplementedError
       end
 
       def stdout(*)
-        NotImplementedError
+        raise NotImplementedError
       end
 
       def stderr(*)
-        NotImplementedError
+        raise NotImplementedError
       end
 
       def close_io(*)
--- a/lib/aruba/processes/in_process.rb
+++ b/lib/aruba/processes/in_process.rb
@@ -103,7 +103,7 @@ module Aruba
       #
       # @param *input [Array of String]
       def write(*input)
-        @stdin.write *input
+        @stdin.write(*input)
       end
 
       # Close io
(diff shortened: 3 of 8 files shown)


lib/aruba/processes/basic_process.rb
```diff
@@ -52,15 +52,15 @@ def write(*)
...
       end
 
       def stdout(*)
-        NotImplementedError
+        raise NotImplementedError
       end
 
       def stderr(*)
-        NotImplementedError
+        raise NotImplementedError
```
olleolleolle: Good fix!

lib/aruba/processes/in_process.rb
```diff
@@ -99,12 +99,11 @@ def stderr(*)
         @stderr.string
       end
 
-      # Write strint to stdin
+      # Write strings to stdin
```
olleolleolle: Oh, again, tidy!

lib/aruba/processes/in_process.rb
```diff
@@ -99,12 +99,11 @@ def stderr(*)
...
       #
-      # @param [String] input
-      #   Write string to stdin in
-      def write(input)
-        @stdin.write input
+      # @param *input [Array of String]
```
olleolleolle:
```suggestion
      # @param *input [Array<String>]
```

spec/aruba/processes/basic_process_spec.rb
```diff
@@ -5,39 +5,41 @@
...
+          @stderr
+        end
+      end.new(stdout, stderr, cmd, exit_timeout, io_wait_timeout, working_directory)
+    end
 
-  describe '#inspect' do
     it 'it shows useful info' do
       expected = /#<#<Class:0x[0-9A-Fa-f]+>:\d+ commandline="foobar": stdout="foo output" stderr="foo error output"/
-      expect(subject.inspect).to match(expected)
+      expect(derived_process.inspect).to match(expected)
```
olleolleolle: This is a neat update to the expect() clauses.

lib/aruba/api/commands.rb
```diff
@@ -186,7 +186,7 @@ def run_command_and_stop(cmd, opts = {})
       def type(input)
         return close_input if input == "\u0004"
 
-        last_command_started.write(input << "\n")
+        last_command_started.write(input, "\n")
```
luke-hill: This probably needs to enumerate something like

```
input.times do |item|
  last_command_started.write(item, "\n")
end
```

Issue #605, opened 2019-03-23
Undeprecate selected parts of the API


mvz:
## Summary

The replacement for some deprecated API methods is rather verbose. We should consider undoing their deprecation or provide better replacements

## Expected Behavior

Aruba provides a simple upgrade path from earlier versions that does not lead to repetition of complex code in users' spec and cucumber files.

## Current Behavior

The replacement for `all_output` is `all_commands.map { |c| c.output }.join("\n")`, and for `in_current_directory` it's `cd(".")`. Especially the first one is rather verbose.

## Possible Solution

We could restore some of these methods to be part of the officially supported API.

## Context & Motivation

The `all_output` example comes from https://github.com/rspec/rspec-core/pull/2609.

mvz:
This was done in #638 and #644.

PR #638, opened 2019-05-31
Undeprecate some stuff

mvz:
## Summary

Restores `#all_output`, `#all_stdout`, `#all_stderr` and `#in_current_directory` API methods to non-deprecated status. Also un-deprecates the current behavior of output-checking cucumber steps.

## Motivation and Context

* These API methods seem to be the deprecated methods most used by downstream projects, and their replacements are more complex.
* The output-checking cucumber steps are currently set to have a breaking change in 1.0.0 and we should instead have a smooth upgrade path that will allow moving from latest 0.14.x to 1.0.0 when the latter is released.
* The replacement cucumber steps in master currently have less useful output, and the current behavior is more intuitive.

## How Has This Been Tested?

Some tests were added but mostly code was just moved.

## Types of changes

- [x] Update documentation

## Checklist:

- [x] I've added (some) tests for my code

### DIFF
--- a/features/03_testing_frameworks/cucumber/steps/command/check_output_of_command.feature
+++ b/features/03_testing_frameworks/cucumber/steps/command/check_output_of_command.feature
@@ -375,7 +375,7 @@ Feature: All output of commands which were executed
     When I run `cucumber`
     Then the features should all pass
 
-  Scenario: Detect output from all processes normal and interactive ones
+  Scenario: Detect combined output from normal and interactive processes
     Given an executable named "bin/aruba-test-cli1" with:
     """
     #!/usr/bin/env bash
--- a/features/03_testing_frameworks/cucumber/steps/command/check_output_of_command.feature
+++ b/features/03_testing_frameworks/cucumber/steps/command/check_output_of_command.feature
@@ -347,7 +347,6 @@ Feature: All output of commands which were executed
     When I run `cucumber`
     Then the features should all pass
 
-  @requires-aruba-version-1
   Scenario: Detect output from all processes
     Given an executable named "bin/aruba-test-cli1" with:
     """bash
@@ -370,34 +369,35 @@ Feature: All output of commands which were executed
         Then the stdout should contain exactly:
         \"\"\"
         This is cli1
-        \"\"\"
-        And the stdout should contain exactly:
-        \"\"\"
         This is cli2
         \"\"\"
     """
     When I run `cucumber`
     Then the features should all pass
 
-  Scenario: Detect output from all processes (deprecated)
+  Scenario: Detect output from all processes normal and interactive ones
     Given an executable named "bin/aruba-test-cli1" with:
-    """bash
+    """
     #!/usr/bin/env bash
-
     echo 'This is cli1'
     """
     And an executable named "bin/aruba-test-cli2" with:
-    """bash
-    #!/usr/bin/env bash
+    """
+    #!/usr/bin/env ruby
 
-    echo 'This is cli2'
+    while input = gets do
+      break if "" == input
+      puts input
+    end
     """
     And a file named "features/output.feature" with:
-    """cucumber
+    """
     Feature: Run command
       Scenario: Run command
         When I run `aruba-test-cli1`
-        When I run `aruba-test-cli2`
+        When I run `aruba-test-cli2` interactively
+        And I type "This is cli2"
+        And I type ""
         Then the stdout should contain exactly:
         \"\"\"
         This is cli1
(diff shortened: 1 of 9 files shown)


mvz:
See #605.

features/03_testing_frameworks/cucumber/steps/command/check_output_of_command.feature
```diff
@@ -370,34 +369,35 @@ Feature: All output of commands which were executed
...
-        And the stdout should contain exactly:
-        \"\"\"
         This is cli2
         \"\"\"
     """
     When I run `cucumber`
     Then the features should all pass
 
-  Scenario: Detect output from all processes (deprecated)
+  Scenario: Detect output from all processes normal and interactive ones
```
luke-hill: normal and interactive processes

lib/aruba/api/deprecated.rb
```diff
@@ -525,11 +509,10 @@ def clean_current_dir(*args, &block)
       def in_current_dir(&block)
         Aruba.platform.deprecated(
           'The use of "in_current_dir" is deprecated.' \
-          ' Use "#cd(\'.\') { }" instead'
+          ' Use "#in_current_directory { }" instead'
```
luke-hill: :+1:

lib/aruba/cucumber/command.rb
```diff
@@ -136,7 +136,30 @@
   expect(last_command_started.output).to have_output_size size.to_i
 end
 
-Then(/^(?:the )?(output|stderr|stdout)(?: from "([^"]*)")? should( not)? contain( exactly)? "([^"]*)"$/) do |channel, cmd, negated, exactly, expected|
+Then(/^(?:the )?(output|stderr|stdout) should( not)? contain( exactly)? "([^"]*)"$/) do |channel, negated, exactly, expected|
```
luke-hill: Whilst this is extremely mutable, actually BDD advocates against this.

It would be better to have twice as many cukes with half the permeability or a quarter e.t.c.

I'm happy for you to defer this, but I'm against introducing more of it

lib/aruba/api/deprecated.rb
```diff
@@ -1017,60 +1000,6 @@ def stderr_from(cmd)
...
-      def all_stdout
-        Aruba.platform.deprecated(
-          'The use of "#all_stdout" is deprecated.' \
-          ' Use `all_commands.map { |c| c.stdout }.join("\n") instead.' \
-          ' If you need to check for some output, use' \
-          ' "expect(all_commands).to included_an_object have_output_on_stdout(/output/)"' \
-          ' instead'
-        )
-
-        process_monitor.all_stdout
```
luke-hill: these delegate to process_monitor but they have been reintroduced in commands.rb as a command_monitor.

I'm assuming this is correct.

mvz:
> Also is this branch being merged into master in some guise so these commands will still exist in 1.0?

Yes, I will make another pull request to re-introduce these in master.

lib/aruba/cucumber/command.rb
mvz: I agree but I'm going to defer this to the 1.0 or post-1.0 release.

mvz: Actually, maybe I should clean that up now.

luke-hill: Up to you, maybe make an issue for tidying up. It's a big refactor task though. So perhaps one best done another time with a clear head/coffee

mvz: Ok, just to make sure I understand correctly, you're saying this should be split so that there are several definitions reading something like:

```ruby
Then(/^the output should contain "([^"]*)"$/ do |expected|
# ...
Then(/^the stderr should contain "([^"]*)"$/ do |expected|
# ...
# etc.
```

Right?

lib/aruba/api/deprecated.rb
mvz: Yes, that's correct. `process_monitor` is also deprecated.

lib/aruba/cucumber/command.rb
luke-hill: 1000%, I'm not saying you need to split on each option. But just removing one permutation would make it so much easier to read / review.

Sometimes unDRYing it makes it so much easier to follow. This stuff is quite tricky.

mvz:
@luke-hill I updated the scenario wording and will defer further clean-up of the cucumber steps to later.

lib/aruba/cucumber/command.rb
mvz: I created #640 for this.

Issue #563, opened 2018-05-07
Run with `@debug`  and `@announce` tag fails with `#to_h` missing


maxmeyer:
## Summary

Running a feature with `@debug`  and `@announce` tag fails with:

~~~
<<-COMMAND
NotImplementedError
COMMAND

undefined method `to_h' for NotImplementedError:Class
Did you mean?  to_s (NoMethodError)
~~~

## Expected Behavior

Command execution should not fail with ruby error.

## Current Behavior

The execution fails with the following error.

~~~
undefined method `to_h' for NotImplementedError:Class
Did you mean?  to_s (NoMethodError)
~~~

## Possible Solution

If I remember correctly, I forgot to add `NotImplementedError#to_h`. Maybe returning just an empty hash is a solution for this. I'm not sure, where we use this class.

## Steps to Reproduce (for bugs)

1. Create feature file
~~~
Feature: My feature
  @debug @announce
  Scenario: Test it
    Given I run `hello world` interactively
    Then the stderr should contain:
    """
    hello world
    """
~~~
2. Create `step_definitions.rb`
~~~
require 'aruba/cucumber'
~~~

3. Run cucumber

## Your Environment

* Version used: 1.0.0.pre.alpha.2
* Operating System and version: Ruby
* Link to your project: internal

luke-hill:
I'll investigate this issue @mvz as it seems easy enough to repro.

Quick question, is this just to be cut against master or against master and still?

mvz:
Looks like it's a problem that only occurs on master, but we'll see.

luke-hill:
I've got it about 90% fixed on master. But interestingly the debug flag masks things in a different way. But I've found where the bug is exposed. Just working out the best way to fix it.

mvz:
Fixed in #664 and #665.

PR #649, opened 2019-06-27
Bugfix/to h cast failing on proc call

luke-hill:
## Summary

Fix https://github.com/cucumber/aruba/issues/563

## Details

The casting was failing when certain situations (nil / blank), was being met. As such just performing a quick type check with a safe cast will fix this small bug.

## How Has This Been Tested?

Additional feature test has been added to check for this with a small ruby script

## Types of changes

- [ ] Update documentation

## Checklist:

Specifically we shouldn't elude to this change being done. Because we don't want people un-knowingly using a type-safe check for hashes, unless they want it (i.e. use ActiveSupport instead)


mvz:
I'm not that keen on adding a monkey-patch on Object for this. Rather, we should investigate where this `NotImplementedError` comes from and fix that.

luke-hill:
This isn't a monkeypatch. It uses a new safe method which doesn't overload anything.

If you want me to move where this is exposed I can do that? But yeh :confused: 

Also the place it's required has also been fixed. (As per the build passing this step)

luke-hill:
Is this GtM @mvz?

mvz:
I'll take another look when temperatures have dropped below 30 here :-).

mvz:
I've taken a look at the feature file and it succeeds with and without the rest of this pull request. I also saw that `@announce` prints the whole Ruby binary, which is confusing to say the least :laughing:.

I'd like to do the following:
* The `command_content` announcer should not dump the entire command if it's a binary (not related to #563 but still relevant because the current behavior makes it super-hard to see what's going on).
* Instead of returning NotImplementedError, we should raise it.
* We should have a scenario that tests the combination of `@announce` and `@debug`, like `features/03_testing_frameworks/cucumber/announce_information_for_troubleshooting.feature` and `features/03_testing_frameworks/cucumber/steps/command/debug_your_command_in_aruba.feature` combined. This should now fail with a NotImplementedError exception, rather than a NoMethodError.
* The announcer should catch NotImplementedError and output a friendly warning that it cannot announce whatever it wanted to announce.

luke-hill:
:man_shrugging: that sounds quite complex and I'm not sure what that involves in doing.

As this PR fixes the issue, is this sufficient to merge in as is. Or does it destroy some other functionaltiy?

The Extra Test I think was to demonstrate some extra functionality aside from this which was helping me fix it.

I don't mind which approach we take here. Just that most of what you've suggested to me sounds a bit complex and might take up your time already which is swamped with stuff to do.

mvz:
I've decided to implement https://github.com/cucumber/aruba/pull/649#issuecomment-516865997 instead (except for the first point).

PR #654, opened 2019-07-01
Fix up approx 20% of the rubocop issues in aruba

luke-hill:
## Summary

Fix up some more rubocop offenses

## Details

Some auto-fixed, some manually fixed. Approx 70 failures have been fixed and the config has been regenerated

## Motivation and Context

Make the codebase better, one offense at a time.

## How Has This Been Tested?

CI

## Types of changes

- [x] Refactoring (cleanup of codebase without changing any existing functionality)
- [ ] Update documentation

## Checklist:


lib/aruba/platforms/command_monitor.rb
```diff
@@ -22,6 +22,10 @@ def nil?
       def method_missing(*)
         fail NoCommandHasBeenStoppedError, 'No last command stopped available'
       end
+
+      def respond_to_missing?(*)
```
mvz: I suppose this is correct for now, but I'd really like to simplify this design.

lib/aruba/platform.rb
```diff
@@ -4,18 +4,10 @@
 # Aruba
 module Aruba
   PLATFORM_MUTEX = Mutex.new
-end
+  PLATFORM = [Platforms::WindowsPlatform, Platforms::UnixPlatform].find(&:match?)
```
mvz: Which cop makes you change `Platform` to `PLATFORM`? I feel this file should at least define `Aruba::Platform`. Renaming may even be a breaking change.

lib/aruba/api/core.rb
```diff
@@ -15,7 +15,7 @@ module Core
 
       # Aruba Runtime
       def aruba
-        @_aruba_runtime ||= Runtime.new
+        @aruba ||= Runtime.new
```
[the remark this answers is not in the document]
luke-hill: I checked everywhere that module was included and there was no mention of `@aruba` anywhere in the entire codebase where that module was included

lib/aruba/platform.rb
luke-hill: Constants should be `SCREAMING_SNAKE_CASE`

mvz: I think we should disable the cop for this particular case, since we're basically setting up a class here.

mvz:
@luke-hill could you please rebase this on current master?

lib/aruba/platform.rb
luke-hill: It seems a proper convoluted way of doing things, but I'll alter this change for now and rebase.

mvz:
@luke-hill I think you merged instead of rebasing? Let's wait for Travis right now and then fix that up.

luke-hill:
I fixed this up and rebased with the latest master, but yeh now we have like 100 failures regarding environment.

I did a merge commit as it's easier than rebasing when your PR is >5 commits. But yeh lord knows what's gone on here. Have we removed something in `master`

lib/aruba/api/core.rb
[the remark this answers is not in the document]
mvz: I think this change may be responsible for a lot of the build failures.

mvz:
I don't have time now, but I can have a deeper look into this later in the week.

luke-hill:
I've fixed up the issues and added a warning about that area. In theory that "shouldn't" have broken it, as we shouldn't be accessing a memo's iVar.

Still it's now all green and slightly better. It's also RtM if you're happy with it.

mvz:
Thanks, @luke-hill!

PR #670, opened 2019-08-21
Update JRuby and build environment on Travis

mvz:
## Summary

Use latest JRuby and Ubuntu in Travis build

## Details

* Remove pinning of Ubuntu distribution
* Always supply JAVA_OPTS to avoid errors/warnings about opens, and update it with latest needed additions.
* Remove outdated override for JRUBY_OPTS

## Motivation and Context

We were testing on 9.2.6 because 9.2.7 broke scenario skipping in Cucumber. Now that 9.2.8 is out we should be able to safely move to that.

## How Has This Been Tested?

Needed options were tested locally by running the specs. Travis will do its thing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-# Use old Ubuntu distribution to avoid JRuby issues
-dist: trusty
 language: ruby
 script: bundle exec rake test
 before_install:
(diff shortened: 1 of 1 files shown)


olleolleolle:
Ah, perhaps we need even more add-opens? 

Or they're being read the wrong way from the YAML.

mvz:
> Or they're being read the wrong way from the YAML.

That's probably it.

olleolleolle:
@mvz The way I have been doing that is to put ALL content on 1 line (extremely unreadable, which is why people avoid this), in the matrix element definition:

```yaml
matrix:
  - rvm: jruby-40000
    env: "JRUBY_ALT1=--debug JRUBY_AMAZE=--add-fnorg --add-more"
```

mvz:
It looks like the `JAVA_OPTS` variable is set correctly: https://travis-ci.org/cucumber/aruba/jobs/574983049#L171, and it contains the proper `--add-opens` for `sun.nio.ch`. :confused:

mvz:
I'm leaving this for now in favor of #671.

PR #671, opened 2019-08-24
Bump JRuby 9.2 build version

mvz:
## Summary

Use latest JRuby in Travis build

## Details

Bump JRuby version to 9.2.8.0 for the Linux and OSX builds.

## Motivation and Context

We were testing on 9.2.6 because 9.2.7 broke scenario skipping in Cucumber. Now that 9.2.8 is out we should be able to safely move to that.

I broke this out from #670 since setting the JAVA_OPTS correctly turns out to be more difficult than I anticipated.

## How Has This Been Tested?

Travis will do its thing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ jobs:
       os: linux
     - rvm: jruby-9.1
       os: linux
-    - rvm: jruby-9.2.6.0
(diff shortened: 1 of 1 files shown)


olleolleolle:
Already green, LGTM

PR #675, opened 2019-12-04
Improve command failure message

deivid-rodriguez:
## Summary

Show useful debugging information when the `:be_successfully_executed` matcher fails.

## Details

I was debugging a cucumber suite, and getting this error:

```
Using the default profile...
.F-----------

(::) failed steps (::)

expected "bundle exec rake part1" to be successfully executed (RSpec::Expectations::ExpectationNotMetError)
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-support-3.9.0/lib/rspec/support.rb:97:in `block in <module:Support>'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-support-3.9.0/lib/rspec/support.rb:106:in `notify_failure'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.9.0/lib/rspec/expectations/fail_with.rb:35:in `fail_with'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.9.0/lib/rspec/expectations/handler.rb:40:in `handle_failure'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.9.0/lib/rspec/expectations/handler.rb:50:in `block in handle_matcher'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.9.0/lib/rspec/expectations/handler.rb:27:in `with_matcher'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.9.0/lib/rspec/expectations/handler.rb:48:in `handle_matcher'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.9.0/lib/rspec/expectations/expectation_target.rb:65:in `to'
/home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/aruba-0.14.12/lib/aruba/api/commands.rb:307:in `run_command_and_stop'
[...]

PR #687, opened 2020-01-01
Improve `expand_path` warnings

deivid-rodriguez:
## Summary

My changes improve the warnings printed when passing an absolute path to aruba's `expand_path`, and also suppress several warnings originated from aruba's own helpers, not from end user's code. 

## Details

This PR improves warnings like this:

```
2020-01-01 18:13:47 +0100 WARN: Using absolute paths in Aruba is not recommended. Set config.allow_absolute_paths = true to silence this warning
```

to instead read as

```
2020-01-01 19:37:06 +0100 WARN: Aruba's `expand_path` method was called with an absolute path at /home/deivid/Code/aruba/lib/aruba/api/filesystem.rb:157, which is not recommended. Change the call to pass a relative path or set config.allow_absolute_paths = true to silence this warning
```

The new warning message is better because it gives information about:

* The method being called that it's printing the warning.
* Where the method that it's printing the warning is being called.
* What is the preferred alternative to fix the warning.

In addition, this method prevents several `aruba` helpers from calling `expand_path` with absolute paths, and thus removes several warnings that would be printed to end users otherwise, without really being in control of fixing them (other than by creating a PR similar to this one :)).

## Motivation and Context

These changes were motivated because I tried to upgrade to latest aruba 1.0.0 pre-release to be able to use #675 in simplecov's specs, and I started seeing several of these warnings. I had no clue on how I should proceed to fix them so I investigated.

I hope this will help other users by either removing "false positive warnings" or by giving better warning messages when they are actually "true positives".

## How Has This Been Tested?

I added several tests, and also made sure that warnings are no longer printed during simplecov specs after these changes.

## Types of changes

- [ ] Update documentation

## Checklist:


features/04_aruba_api/filesystem/check_if_path_is_directory.feature
```diff
@@ -37,6 +37,27 @@ Feature: Check existence of directories
     When I run `rspec`
     Then the specs should all pass
 
+  Scenario: Does not warn when passed an absolute path
```
mvz: I think it should warn about this.

features/04_aruba_api/filesystem/check_if_path_is_file.feature
```diff
@@ -38,6 +38,27 @@ Feature: Check existence of files
     When I run `rspec`
     Then the specs should all pass
 
+  Scenario: Does not warn when passed an absolute path
```
mvz: I also think this should warn.

deivid-rodriguez:
@mvz Since you agree with improving the messages, I removed the change fixing what I think are false positive warnings. Let's discuss the improvements in the warning message first, and then I'll open a separate PR to discuss whether all file helpers should warn or not when passed absolute paths. Sounds good?

mvz:
Looks good, thanks!

mvz:
And thanks for trying out the 1.0.0 pre-release, @deivid-rodriguez!

deivid-rodriguez:
No problem!

I investigated a bit more the warnings I'm getting and I tracked them down to using the `The following files should exist:` step with absolute paths, which ends up calling the `be_an_existing_file` matcher with the same absolute paths, which ends up calling `expand_path` with the same absolute paths.

This seems currently hard to track down (even with the improvements in this PR) because it's not the end user code that calls `expand_path` but internal aruba's helpers, steps or matchers. For example, the warnings I get look like this:

```console
$ bundle exec cucumber
Using the default profile...
............................2020-01-02 11:53:16 +0100 WARN: Using absolute paths in aruba at /home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/aruba-1.0.0.pre.alpha.5/lib/aruba/api/filesystem.rb:38. Using absolute paths in Aruba is not recommended. Set config.allow_absolute_paths = true to silence this warning
2020-01-02 11:53:16 +0100 WARN: Using absolute paths in aruba at /home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/aruba-1.0.0.pre.alpha.5/lib/aruba/api/filesystem.rb:30. Using absolute paths in Aruba is not recommended. Set config.allow_absolute_paths = true to silence this warning
2020-01-02 11:53:16 +0100 WARN: Using absolute paths in aruba at /home/deivid/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/aruba-1.0.0.pre.alpha.5/lib/aruba/api/filesystem.rb:30. Using absolute paths in Aruba is not recommended. Set config.allow_absolute_paths = true to silence this warning
..................................................................................................................................................................................................................................................................

55 scenarios (55 passed)
286 steps (286 passed)
0m56.172s
```

Better than before, because it gives me a place to start looking, but still not very user friendly.

If you really want to fix this for end users, I think that would involve showing proper warnings with proper user code locations for every end user `file` method, step or matcher being called with an absolute path argument, and then silencing any other warnings that might be triggered internally during its execution.

But, I would honestly remove the warning. The original issue where the problem being fixed here was reported is https://github.com/cucumber/aruba/issues/478 (not sure if you got other related problems reported over time). The scenario mentioned in the ticket is exactly the same scenario in simplecov that it's throwing the warning for me now. In this case, I consider our scenario perfectly fine, so my solution (now that I know what's happening) will be the `config.allow_absolute_paths` thing.

But still, if the only problem you've been reported is considered as a false positive, it seems plausible to think that many other users will consider their cases as false positives to, so it seems questionable to warn everybody's code in exchange for fixing "potential bugs" nobody has reported.

Hope this helps!

deivid-rodriguez:
Also, if you go the "warn everything" way, the `expand_path` documentation should probably be updated, since it currently shows an example with an absolute path:

https://github.com/cucumber/aruba/blob/c2fed660529b163faba41bd4322a4f3b01ec0c78/lib/aruba/api/core.rb#L129-L133

without mentioning that it warns.

mvz:
I opened #691 to improve the message further.

PR #701, opened 2020-01-23
Update development dependencies

mvz:
## Summary

Update dependencies to their latest version

## Details

Updates license finder and rubocop. Adjust license finder configuration because the whitelist option was changed to permit.

## Motivation and Context

Keeps things up-to-date.

## How Has This Been Tested?

Travis will tell.


olleolleolle:
> # Scenario: Sending signal to the command started last

That scenario, on macOS, kind of flails a little

mvz:
> > `# Scenario: Sending signal to the command started last`
> 
> That scenario, on macOS, kind of flails a little

Yes, it's very annoying and hard to debug. We need someone with a MacOS system to debug it, if they can reproduce it.

PR #702, opened 2020-01-25
Find relative commands from the current Aruba directory

### LINKED ISSUE #552: Executing relative paths, oh my!
## Summary

Aruba looks for executables by relative path starting from the current directory, not from the working directory. This means a feature cannot create an executable and immediately run it. There are some corner cases that we need to think about

## Expected Behavior

A scenario like this should work:
```
    Feature: Run relative command

    Scenario: Run command
      Given an executable named "bin/local_cli" with:
      """
      #!/bin/bash
      exit 0
      """
      When I successfully run `bin/local_cli`
```

## Current Behavior

This fails with `Command "bin/local_cli" not found in PATH-variable`. This is because Aruba tries to find `bin/local_cli` in what Ruby thinks is the current directory (usually the project directory), rather than `aruba.current_directory`.

## Possible Solution

The radical solution is to look for relative executables in `aruba.current_directory`. I think this is the correct way, because:

* Executables that are part of the project under test are already available through the PATH, so using `bin/foo` for that does not make sense.
* `aruba.current_directory` represents where we 'are', during the running of a feature, and typing `bin/local_cli` there would resolve starting from that point.

**One problem may be projects that rely on the current broken behavior.**

## Context & Motivation

I'm trying to upgrade the aruba dependency for `rspec-rails`. That gem still depends on version `~> 0.5.4.`, probably because of this issue. My conclusion is that this is something that got broken between 0.5.4 and 0.14.x.

mvz:
## Summary

Find relative commands from the current Aruba directory

## Details

In an Arabu spec or scenario, files are created relative to the current Aruba directory. When creating an executable in that way, it should be possible to run that executable using the same relative path.

This change means that executables in the project under test can no longer be find using the relative path from the project directory. They should be added to the path. This is done automatically by Bundler.

Technically this is accomplished by doing the command lookup inside an `in_current_directory` block.

## Motivation and Context

Fixes #552. 

## How Has This Been Tested?

Specs and cukes were added.

## Types of changes

- [x] Update documentation

## Checklist:

### DIFF
--- a/features/04_aruba_api/command/run_command.feature
+++ b/features/04_aruba_api/command/run_command.feature
@@ -23,7 +23,7 @@ Feature: Run command
   Background:
     Given I use a fixture named "cli-app"
 
-  Scenario: Existing executable
+  Scenario: Executable in the project's path 
     Given an executable named "bin/aruba-test-cli" with:
     """bash
     #!/bin/bash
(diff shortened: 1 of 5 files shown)


features/04_aruba_api/command/run_command.feature
```diff
@@ -41,19 +41,24 @@ Feature: Run command
...
 
     RSpec.describe 'Run command', type: :aruba do
-      before(:each) { run_command('bin/aruba-test-cli') }
-      it { expect(last_command_started).to be_successfully_executed }
+      before do
+        write_file 'bin/aruba-test-cli', <<~BASH
+          #!/bin/bash
+          exit 0
+        BASH
+        chmod 0x755, 'bin/aruba-test-cli'
```
olleolleolle: This says 0x755 where the other one says 0o755, is there a difference by choice? Or by accident?

mvz: Seems to be a RuboCop correction. I'll equalize these.

mvz: Nope, it was just wrong :slightly_smiling_face:. It should be `0o755` or `0755`.

olleolleolle:
Yay hooray!

Issue #720, opened 2020-07-14
Drop `git` from gemspec


utkarsh2102:
Hi,

Thanks for working on this! :heart: 
However, while maintaining this in Debian, we found that this library relies on `git` to list the files (which could be done via pure Ruby alternative):

https://github.com/cucumber/aruba/blob/886490e037394223149c9b3f3c510e8c4a4900c5/aruba.gemspec#L48-L50

I'd be happy to raise a PR if you concur in dropping `git` from gemspec :smile: 
Let me know what you think!?

PR #721, opened 2020-07-16
Drop `git ls-files` in gemspec

utkarsh2102:
Hi,

Thanks for working on this! :heart: 
However, while maintaining this in Debian, we found that this library relies on `git` to list the files which could be done via pure Ruby alternative -- which is what this PR does.

As an addition, this PR makes sure that this gem only ships the required files to the end-users and not other things which are not needed by them! :rocket: 

Also, added `rubocop-packaging` as a `development_dependency` which will ensure the best practices.
Here's what it shows:

```
Inspecting 172 files
...C..........C.........C...........
....................................
....................................
....................................
............................

Offenses:

aruba.gemspec:48:16: C: Packaging/GemspecGit: Avoid using git to produce lists of files.
Downstreams often need to build your package in an environment that does not have git (on purpose).
Use some pure Ruby alternative, like Dir or Dir.glob.
  spec.files = `git ls-files -z`.split("\x0").reject do |f|
               ^^^^^^^^^^^^^^^^^
fixtures/cli-app/cli-app.gemspec:18:24: C: Packaging/GemspecGit: Avoid using git to produce lists of files.
Downstreams often need to build your package in an environment that does not have git (on purpose).
Use some pure Ruby alternative, like Dir or Dir.glob.
  spec.files         = `git ls-files -z`.split("\x0")
                       ^^^^^^^^^^^^^^^^^
fixtures/empty-app/cli-app.gemspec:18:24: C: Packaging/GemspecGit: Avoid using git to produce lists of files.
Downstreams often need to build your package in an environment that does not have git (on purpose).
Use some pure Ruby alternative, like Dir or Dir.glob.
  spec.files         = `git ls-files -z`.split("\x0")
                       ^^^^^^^^^^^^^^^^^

172 files inspected, 3 offenses detected
```

And this PR fixes the same, which helps us in shipping this in Debian! :tada: 
Hope this would make sense and you'll be open to this change :100: 

Signed-off-by: Utkarsh Gupta <<utkarsh@debian.org>>

---

CC: @luke-hill 
Closes: #720

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -1,6 +1,7 @@
 lib = ::File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'aruba/version'
+require 'rake/file_list'
 
 Gem::Specification.new do |spec|
   spec.name        = 'aruba'
@@ -46,7 +47,8 @@ Gem::Specification.new do |spec|
   spec.rubygems_version = '>= 1.6.1'
   spec.required_ruby_version = '>= 2.4'
 
-  spec.files = Dir['{docs,fixtures,lib}/**/*', 'CHANGELOG.md', 'LICENSE', 'README.md']
+  spec.files = Rake::FileList['{docs,fixtures,lib}/**/*', 'CHANGELOG.md', \
+                              'LICENSE', 'README.md'].exclude(*File.read('.gitignore').split)
 
   spec.executables      = ['exe/aruba']
   spec.rdoc_options     = ['--charset', 'UTF-8', '--main', 'README.md']
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -1,4 +1,4 @@
-lib = ::File.expand_path('../lib', __FILE__)
+lib = ::File.expand_path('lib', __dir__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'aruba/version'
 require 'rake/file_list'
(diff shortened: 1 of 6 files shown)


aruba.gemspec
```diff
@@ -45,11 +46,9 @@ Gem::Specification.new do |spec|
   spec.rubygems_version = '>= 1.6.1'
   spec.required_ruby_version = '>= 2.4'
 
-  spec.files = `git ls-files -z`.split("\x0").reject do |f|
-    f.match(%r{^(test|spec|features)/})
-  end
+  spec.files = Dir['{docs,fixtures,lib}/**/*', 'CHANGELOG.md', 'LICENSE', 'README.md']
```
mvz, line 49: Please add a `.exclude(*File.read('.gitignore').split)` like you did in for cucumber-rails and gir_fii.

utkarsh2102: Aw, crap. Right!

utkarsh2102: Should've done earlier but done now!
Thanks for the review \o/

aruba.gemspec
```diff
@@ -46,7 +46,8 @@ Gem::Specification.new do |spec|
   spec.rubygems_version = '>= 1.6.1'
   spec.required_ruby_version = '>= 2.4'
 
-  spec.files = Dir['{docs,fixtures,lib}/**/*', 'CHANGELOG.md', 'LICENSE', 'README.md']
+  spec.files = Rake::FileList['{docs,fixtures,lib}/**/*', 'CHANGELOG.md', \
```
olleolleolle, line 49: This PR now does not have the explicit `require 'rake/file_list'` - is that significant?

utkarsh2102: Eeks. Thanks, @olleolleolle. Fixed! :heart:

PR #725, opened 2020-08-19
Use manifest file to specify set of files to include in gem

mvz:
## Summary

Stop dynamically collecting files to include in the gem and instead specify (most of) them in a manifest file

## Details

Use rake-manifest gem to generate and check a `Manifest.txt` file, and read that file in the gemspec to set the list of files to include in the gem. Also, run manifest checking task before gem building task.

## Motivation and Context

- Using git to list files is problematic (#721)
- The solution from #721 does not accurately exclude files
- The solution from #721 needs parts of rake that may not be installed when e.g. Bundler is run to install dependencies, so just running `bundle install` to fetch dependencies from a fresh checkout doesn't work, leading to build failures on TravisCI

## How Has This Been Tested?

- Ran rake build
- Travis

## Types of changes


aruba.gemspec
```diff
@@ -36,19 +33,19 @@ Gem::Specification.new do |spec|
...
+  spec.add_development_dependency 'rubocop-rspec', '~> 1.42.0'
   spec.add_development_dependency 'simplecov', '~> 0.18.0'
   spec.add_development_dependency 'yard-junk', '~> 0.0.7'
 
   spec.rubygems_version = '>= 1.6.1'
   spec.required_ruby_version = '>= 2.4'
 
-  spec.files = Rake::FileList['{docs,fixtures,lib}/**/*', 'CHANGELOG.md', \
-                              'LICENSE', 'README.md'].exclude(*File.read('.gitignore').split)
+  spec.files = File.read('Manifest.txt').split
```
olleolleolle, line 48: Curious: can this use https://rubyapi.org/2.7/o/io#method-c-readlines ? With chomp: true I guess.

olleolleolle (commented):
I made a premature hello-there inline comment.

aruba.gemspec
mvz: Yes! I didn't know about `chomp: true`; much nicer!

PR #732, opened 2020-09-27
Update development infrastructure

mvz:
## Summary

Update to the development infrastructure to reduce repetitive work and waiting

## Details

- Configure Dependabot, so dependency updates are done mostly automatically
- Restructure Travis build tasks to reduce waiting

## Motivation and Context

Making pull requests just to keep the dependencies updated is boring, so let's make the cloud do it. Also, the sooner Travis is done building the better.

## How Has This Been Tested?

We'll know if it works once Travis does its thing

## Types of changes

- Developer experience improvement


olleolleolle:
Awesome

Issue #699, opened 2020-01-09
Help Request: Testing a separate ruby project within another


PragTob:
Hello there and thanks for aruba :green_heart: 

My question/scenario might be a bit weird but I'll try either way. Code ref: https://github.com/colszowka/simplecov/pull/814 (PR where I stumbled upon this problem trying to implement something)

This is a request for "How do I do this? Can I do this?" _not_ "please change the way you're doing things"

## Versions

Ruby: 2.7
Aruba: 0.14.14

## Context

SimpleCov. We need to make sure SimpleCov works in many different contexts as such we have a "fake project" that we run tests in and then want to make sure the returned coverage results for running tests there are right.

To do this, we have a slightly awkward `Before` hook replacing the project contents with the faked project of our choice (note: I didn't write this and in general, better ways of doing this are welcome :D )

```ruby
Before do
  this_dir = File.dirname(__FILE__)

  # Clean up and create blank state for fake project
  cd(".") do
    FileUtils.rm_rf "project"
    FileUtils.cp_r File.join(this_dir, "../../spec/faked_project/"), "project"
  end

  step 'I cd to "project"'
end
```

(adapted in the current PR to support different projects)

## Problem

To the best of my knowledge/debugging aruba executes command like `bundle` or `bundle exec parallel_rspec spec` in the context of the bundle of the original app (simplecov, not our faked project). Commands are run via "When I successfully run `cmd`"

output captured from running the cucumber tests (from the PR):

<details>

```
tobi@speedy:~/github/simplecov(more-test-projects)$ be cucumber features/parallel_tests.feature 
Using the default profile...

Using rake 13.0.1
Using public_suffix 4.0.2
Using addressable 2.7.0
Using mini_mime 1.0.2
Using mini_portile2 2.4.0
Using nokogiri 1.10.7
Using rack 2.0.8
Using rack-test 1.1.0
Using regexp_parser 1.6.0
Using xpath 3.2.0
Using capybara 3.30.0
Using websocket-extensions 0.1.4
Using websocket-driver 0.7.1
Using apparition 0.4.0
Using childprocess 3.0.0
Using contracts 0.16.0
Using builder 3.2.4
Using backports 3.15.0
Using cucumber-tag_expressions 1.1.1
Using gherkin 5.1.0
Using cucumber-core 3.2.1
Using cucumber-expressions 6.0.1
Using cucumber-wire 0.0.1
Using diff-lcs 1.3
Using multi_json 1.14.1
[...]

PR #749, opened 2020-11-28
Disable bundler using bundler tooling (0-14-stable)

mvz:
## Summary

Use Bundler's own tooling to break out of the Bundler environment, instead of just unsetting a bunch of environment variables.

## Details

Use `Bunder.with_unbundled_env` to inform us how manipulate the environment instead of guessing ourselves: We fetch the environment inside the `Bunder.with_unbundled_env` block and record the changes that were applied. Then we manipulate Aruba's environment to have the same changes.

## Motivation and Context

#699.

## How Has This Been Tested?

Let's see what TravisCI has to say ...

## Types of changes


olleolleolle (approved):
Thanks for showing how to do this, @mvz - looks simpler, and built into Bundler.

Let's get this merged!

PR #755, opened 2020-12-04
Build 0-14-stable branch with GitHub Actions

mvz:
## Summary

Like on master, move to GitHub Actions on this old stable branch as well.

*Update:* Use GitHub Actions for newer Rubies, and TravisCI for the older ones.

## Types of changes

- Update of development tooling


mvz:
This won't work since these old Rubies are not supported.

PR #757, opened 2020-12-04
Loosen dependency on simplecov

mvz:
## Summary

Loosen dependency on simplecov

## Motivation and Context

The dependency of simplecov is a bit hard for dependabot to update automatically.

## How Has This Been Tested?

Updated the bundle and ran the specs.

## Types of changes

- Update development infrastructure


olleolleolle (approved):
👍 LGTM

PR #763, opened 2020-12-25
Run tests on Ruby 3.0 in CI

mvz:
## Summary

Add Ruby 3.0 to the list of Rubies to test on with GitHub Actions

## Motivation and Context

Ruby 3.0 is out!


olleolleolle:
Passes on Ubuntu!

The failure seems to be macos.

PR #775, opened 2021-01-09
Give stderr output example a little more time

mvz:
## Summary

Increase io_wait_timeout a little larger so it's not exceeded during
MacOS cucumber runs in CI.

## Details

Increase io_wait_timeout in one cucumber scenario. Also remove pointless
setting of that value outside of the fixture app directory.

## Motivation and Context

Somehow it seems there's always at lease one MacOS build failing.

## How Has This Been Tested?

I ran the scenarios locally but that doesn't really tell much.

## Types of changes


mvz:
It's green :green_heart: :green_heart: :green_heart: !

PR #780, opened 2021-01-24
Add step and API to add whole lines to a file

### LINKED ISSUE #662: The 'append to file' steps should add newlines
## Summary

Using `append to file` cuke, shouldn't cause
```
syntax error, unexpected end-of-input, expecting keyword_end (SyntaxError)
```

## Expected Behavior

It wouldn't cause the error above

## Current Behavior

It does cause the error above. Code similar to this can be produced

```
def path_to(page_name)
    self.send(page_name.split(' ').push('path').join('_').to_sym)
  end

  Then('I should see {string}') do |text|
    expect(page).to have_content(text)
  endWhen('I go to the session establish page') do # This is the problematic line
  visit path_to('session establish')
end
```

## Possible Solution

Check for newline characters. If one isn't found, add one.
Adding an extra newline even in the worst circumstance, is just a bad style, but not adding one can occassionally cause issues concatenating `end` and `Given` or `When` for example.

1. Use cuke `write to file`
2. Use cuke `append to file` in the same scenario with an identical path
3.
4.

## Context & Motivation

This is present in both scenario and background -> scenario contexts.

I've tagged this as a bug because it does feel something that shouldn't happen

mvz:
## Summary

Add step and API to add whole lines to a file

## Details

Due to how multi-line string literals work in Gherkin, it is easy to accidentally have fewer newlines than intended when building up a file using append_to_file.

To mitigate this while still allowing building up files without inserted newlines, this change introduces a new API method `#append_lines_to_file` and a corresponding step definition, that will ensure a newline is present between the original content and the added content.

## Motivation and Context

Fixes #662.

## How Has This Been Tested?

A new Cucumber scenario and RSpec specs were added.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:


features/03_testing_frameworks/cucumber/steps/filesystem/append_to_file.feature
```diff
@@ -26,6 +26,28 @@ Feature: Append content to file
     When I run `cucumber`
     Then the features should all pass
 
+  Scenario: Append whole lines to a existing file
```
olleolleolle, line 29: An existing

olleolleolle (approved):
Looks good! Found an easy typo.

olleolleolle (approved):
Wanted to say YAY you have such drive!

spec/aruba/api/filesystem_spec.rb
```diff
@@ -4,6 +4,23 @@
 RSpec.describe Aruba::Api::Filesystem do
   include_context "uses aruba API"
 
+  describe "#append_lines_to_file" do
+    let(:path) { @file_path }
```
luke-hill, line 8: pop these as top level? Seem to be used in nearly all of the examples (I scrolled down past L40)

luke-hill (approved):
Minor refactor, not needed here. Do if you want.

spec/aruba/api/filesystem_spec.rb
mvz: That makes sense, yes.

Issue #726, opened 2020-08-21
The append_to_file method should provide feedback if file does not exist


luke-hill:
## Summary

When using `When I append to file "...." with: ` one of the following should happen if the file doesn't exist.
- A warning should be thrown
- An error should be thrown

## Current Behavior

It silently does nothing

## Steps to Reproduce (for bugs)

Self explanatory

## Context & Motivation

Was fixing some stuff in cucumber-rails. Found this issue when writing some new features.

mvz:
> It silently does nothing

What do you mean by this? From the code, it seems it silently creates the file. Is that the behavior you're seeing?

mvz:
This is officially documented bad behavior: https://github.com/cucumber/aruba/blob/996d631d8baa6b447aef4d2d6543b482060b381e/features/03_testing_frameworks/cucumber/steps/filesystem/append_to_file.feature#L29-L43

Steps to resolve this:

- In 1.x: Deprecate silent creation of the file
- In 2.0: Remove silent creation of the file

luke-hill:
Sounds perfect. Good solution, definite thumbs up from me.

PR #781, opened 2021-01-24
Deprecate file creation when using append_to_file

mvz:
## Summary

Deprecate file creation when using `#append_to_file`

## Details

This change makes `#append_to_file` check if the file exists. If it does not, a deprecation message is shown and the file is created. In Aruba 2.0, the deprecated behavior will be removed.

## Motivation and Context

The #append_to_file method creates the file if it does not exist. In such cases, #create_file should be used instead.

See #726.

## How Has This Been Tested?

The relevant cucumber scenario has been amended to check for the presence of the deprecation message.

## Types of changes

- Deprecation
- Documentation update

## Checklist:


olleolleolle (approved):
A sensible approach.

Linting warns of a too-long module. Oh well.

features/03_testing_frameworks/cucumber/steps/filesystem/append_to_file.feature
```diff
@@ -26,7 +26,7 @@ Feature: Append content to file
     When I run `cucumber`
     Then the features should all pass
 
-  Scenario: Append to a non-existing file
+  Scenario: Append to a non-existing file (deprecated)
```
luke-hill, line 29: Question (not a sticking point)

Is this the format we'll be doing in features for deprecated stuff? We can copy to other repos if desired.

luke-hill (approved):

features/03_testing_frameworks/cucumber/steps/filesystem/append_to_file.feature
mvz: I think it's clear this way, but maybe there are other options.

PR #788, opened 2021-02-26
Simplify Docker setup

### LINKED ISSUE #682: Update docker development support
## Summary

The current Dockerfile is outdated and the current support for development with Docker is too complex.

## Expected Behavior

* The Dockerfile is up-to-date and provides a Ruby version that is supported by Aruba
* The Dockerfile is easy to update for Aruba developers even if they do not use Docker every day
* The Dockerfile and docker-compose.yml are well documented and do not contain any mystery parts
* Developing with docker uses standard docker command line tools
* The Dockerfile uses an existing image provided by the Ruby community

## Current Behavior

* The Dockerfile is long and complex
* The Dockerfile builds its own Ruby
* The Dockerfile provides Ruby 2.3
* The Rakefile provides a complicated set of tasks to run tests with docker
* There is a docker-compose file, but it is only used to provide settings for the rake task, which thus duplicates docker-compose functionality

## Possible Solution

* Clean up Dockerfile, basing it on the `ruby:2.4` base image
* Provide clear instructions on how to use docker and/or docker-compose directly
* Remove docker-related rake tasks

## Context & Motivation

We're dropping support for Ruby 2.3 and I just noticed it is what is provided by the current Dockerfile.

mvz:
## Summary

Simplify Docker setup

## Details

- Use existing Ruby image as base to avoid need to use rvm
- Replace docker-compose configuration with instructions in `Dockerfile`
- Remove docker-related rake tasks and their supporting infrastructure
- Remove `bin/test` script
- Update instructions in `CONTRIBUTING.md` accordingly

## Motivation and Context

This makes the `Dockerfile` more maintainable by non-docker-experts like myself. In turn, this opens the possibility of testing on Ruby versions that are harder to install, like Ruby 2.3.

Fixes #682.

## How Has This Been Tested?

I successfully ran the test suite by following the instructions in the `Dockerfile` and then running `bundle exec rake test` inside the Docker container.

## Types of changes

## Checklist:


mvz:
I should also update `CONTRIBUTING.md`.

Dockerfile
```diff
@@ -1,67 +1,43 @@
...
+
+RUN mkdir /aruba
+RUN chown aruba:aruba /aruba
+
+# Run the rest of the steps as non-root
+USER aruba
+WORKDIR /aruba
+
+# Ensure Bundler 2.x is installed
+RUN gem update bundler
```
olleolleolle, line 32: `gem install bundler` would do the thing, too?

Dockerfile
```diff
@@ -1,67 +1,43 @@
...
+
+# Add just the files needed for running bundle install
+ADD Gemfile aruba.gemspec Manifest.txt /aruba/
+ADD lib/aruba/version.rb /aruba/lib/aruba/version.rb
+ADD exe/aruba /aruba/exe/aruba
```
olleolleolle, line 37: Interested: If we used `\` line-continuations, would a single `ADD` do the right thing?

mvz: I'm open to that but I'm not sure how that works when the files live in different directories.

Dockerfile
mvz: Probably, but for now I think this makes it clearer that we don't consider the Bundler version shipped in the base image new enough.

Dockerfile
olleolleolle: You're right!

PR #787, opened 2021-02-26
Support for upcoming RSpec 4.0

mvz:
## Summary

Several changes to allow use of Aruba with RSpec 4, and use of Aruba 1.0 to test RSpec 4.

## Details

- Loosen dependency on rspec-expectations so Aruba users can upgrade to RSpec 4 when it comes out.
- Add back support for Ruby 2.3 to support RSpec's (tentative) wish to support that Ruby version for the 4.x series.

I may split this up later once it becomes clearer whether testing RSpec with cucumber 2, 3 or 4 is possible (currently each of these versions has some problem).

## Motivation and Context

See https://github.com/rspec/rspec/issues/61#issuecomment-786505319 and subsequent comments for some discussion.

## How Has This Been Tested?

CI for now. Hopefully the new 2.3 build targets will behave.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Internal change (refactoring, test improvements, developer experience or update of dependencies)

## Checklist:

- [ ] I've added tests for my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #790, opened 2021-02-28
Restore support for Ruby 2.3

mvz:
## Summary

Lowers the required ruby version to 2.3 and makes the necessary changes to the code to make that work.

## Details

- Lower ruby versions in gemspec, rubocop configuration and Dockerfile
- Use most recent version of rubocop that still supports Ruby 2.3
- Undo several code changes that broke 2.3 compatibility

## Motivation and Context

This will allow RSpec 4 to use Aruba 1 for its cucumber scenario's, rather than staying on super-ancient 0.14.x. See also https://github.com/rspec/rspec/issues/61#issuecomment-786505319 and #787.

## How Has This Been Tested?

This was split off from #787 which was already tested in CI. Using the Dockerfile allows easy testing in a Ruby 2.3 environment.

## Types of changes

- New feature (non-breaking change which adds functionality)
- Documentation update

## Checklist:


mvz:
Current plans for RSpec are to keep the current cucumber + aruba dependencies and only upgrade after the 4.0 release. This means this PR can be on hold until then.

olleolleolle:
Super to have the context for the plan, as well as the plan itself!

mvz:
It is no longer realistic from a maintenance standpoint for Aruba 1.x or upcoming 2.x to support Ruby 2.3. RSpec will have to stay on 0.14.x until they drop support for 2.3.

PR #794, opened 2021-03-21
Improve linting tasks

mvz:
## Summary

Some improvements to the linting tasks

## Details

- Move `license_finder` out of the bundle and lint tasks and install it separately in CI

Additionally:

- Simplify rubocop and yard:junk task definitions
- Provide kramdown dependency so yard:junk doesn't error out

## Motivation and Context

License finder has a lot of dependencies but is slow to update them. We also don't need to depend on any particular version.

## How Has This Been Tested?

I rank the lint task with the new definition.

## Types of changes

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -55,7 +55,7 @@ jobs:
     - name: Run cukes
       run: bundle exec rake cucumber
 
-  lint:
+  checks:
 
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
This simplifies configuration, and takes control of what CI does in a more  direct way!

Kudos!

Issue #729, opened 2020-08-21
Test with multiple versions of Cucumber


mvz:
## Summary

Add some mechanism to test that Aruba still works with older versions of Cucumber that are officially supported.

## Expected Behavior

At least the combination of Aruba with Cucumber 2.4 is tested on TravisCI.

## Current Behavior

Tests are run with the latest Cucumber version supported on each Ruby version. This means versions older than Cucumber 4 are not tested.

## Possible Solution

Use the appraisal gem to test with different combinations of dependencies.

## Context & Motivation

I'd like to start using more modern ways of defining the steps, and also improve step wording, but I'm afraid I'll break support for older Cucumber versions.

mvz:
The upcoming end of support for Ruby 2.5 may be a good time to increase the minimum versions of Aruba's dependencies. This should mitigate this problem as well.

mvz:
Adding Appraisals may also help in testing with upcoming Cucumber releases. See #801.

aurelien-reeves:
Short summary of the discussion we had in #801:

We keep dependency to `cucumber` bounded to the actual version of `cucumber`.

That is source of issues when working on releasing new versions of `cucumber`. We had a really hard time releasing `cucumber` version 6 because of circular dependencies: `cucumber` depends on `cucumber-wire` which depends on `aruba` which depends on `cucumber`.

It has been suggested to remove the upper-bound dependency version. That way, releasing new version of `cucumber` would be easier.

It has also been suggested to have some tests for `aruba` using `cucumber`'s master branch.

This is still an open discussion.

PR #801, opened 2021-04-12
Allow newer versions of Cucumber

mattwynne:
This allows us to test / release new versions of Cucumber without first having to come here to make a new release of Arubua.
    
I think the only reason to cap the max version would be if we knew there was a backwards-compatibility problem with a release of Cucumber, but there isn't, so I think we can safely just remove it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG][2].
 
 ## [Unreleased]
 
-* Update dependencies to cucumber to allow depends on future version
+* Update dependencies to cucumber to allow working with incoming major versions
   ([#801] by [mattwynne])
 
 ## [1.0.4] / 2021-01-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ This project adheres to [Semantic Versioning][1].
 
 This document is formatted according to the principles of [Keep A CHANGELOG][2].
 
+## [Unreleased]
+
+* Update dependencies to cucumber to allow depends on future version
+  ([#801] by [mattwynne])
+
 ## [1.0.4] / 2021-01-04
 
 * Update rubocop and fix new offenses (various pull requests)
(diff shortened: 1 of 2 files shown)


mvz:
I'm not a big fan of open-ended dependencies: If Cucumber releases a new major version, it _may_ break Aruba. On the other hand, since Cucumber uses Aruba for its own tests, that is not very likely at the moment.

What would be other possible ways to solve the problem?

aurelien-reeves:
> 
> 
> I'm not a big fan of open-ended dependencies: If Cucumber releases a new major version, it _may_ break Aruba. On the other hand, since Cucumber uses Aruba for its own tests, that is not very likely at the moment.
> 
> What would be other possible ways to solve the problem?

Actually, at the moment, we cannot release a new version of cucumber because of aruba.
We could, but we cannot execute tests before releasing it.

I share the thinking of @mattwynne: while everything is working fine, having an open-ended dependency is not an issue.
If some day cucumber break aruba, we can release a patched version of aruba to restrict the usage of some specific version of cucumber, then work on a minor or major new release to fix the issue we would face with that new cucumber.

What do you think?

mvz:
> while everything is working fine, having an open-ended dependency is not an issue.

Of course. I have an issue with the case where not everything is fine :slightly_smiling_face:. Maybe I can bump the max version to "< 7.0" and perhaps even test with cucumber's master branch?

aurelien-reeves:
> 
> 
> > while everything is working fine, having an open-ended dependency is not an issue.
> 
> Of course. I have an issue with the case where not everything is fine 🙂. Maybe I can bump the max version to "< 7.0" and perhaps even test with cucumber's master branch?

Bumping to version < 7.0 would actually result in the exact same risk, but less future-proof:

Cucumber version 6.0 actually does not exists yet. Maybe aruba is not compatible with it, we just don't know yet, and we won't until cucumber@6 is released.

Having some tests which uses cucumber's master branch would help us to detect issues before any releases.

mvz (approved):

aurelien-reeves:
In order to be able to move-on with our actual release of cucumber-wire and cucumber, we have decided to keep an upper-bound version number here and to continue the discussion there: https://github.com/cucumber/aruba/issues/729

PR #802, opened 2021-04-14
Stabilization of windows ci build

aurelien-reeves:


mvz:
Thanks @aurelien-reeves!

PR #804, opened 2021-05-10
Loosen dependency on the contracts gem

mvz:
## Summary

Loosen dependency on the contracts gem to allow version 0.17.

## Motivation and Context

Keep things up-to-date. We still allow 0.16 because 0.17 requires Ruby 3.0.

## How Has This Been Tested?

CI will check it.

## Types of changes


olleolleolle (approved):
👍

PR #810, opened 2021-06-02
bugfix: Undefined module/class `Bundler` when running cucumber-rails from RubyMine

luke-hill:
Bundler is a runtime dependency inside `aruba/api/bundler` - So require it as so

## Summary

Fix gemspec to pull in bundler as a requirement of using the gem (As it's needed to run some methods)

## Details

Self-explanatory

## Motivation and Context

When running cucumber-rails in a test/valid context, this throws an error when consuming aruba helpers

## How Has This Been Tested?

CI / Manual
## Types of changes

- [ ] Documentation update

## Checklist:

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -21,12 +21,12 @@ Gem::Specification.new do |spec|
     "source_code_uri" => "https://github.com/cucumber/aruba"
   }
 
+  spec.add_runtime_dependency "bundler", [">= 1.17", "< 3.0"]
   spec.add_runtime_dependency "childprocess", [">= 2.0", "< 5.0"]
   spec.add_runtime_dependency "contracts", "~> 0.16.0"
   spec.add_runtime_dependency "cucumber", [">= 2.4", "< 6.0"]
   spec.add_runtime_dependency "rspec-expectations", "~> 3.4"
   spec.add_runtime_dependency "thor", "~> 1.0"
-  spec.add_runtime_dependency "bundler", "~> 2.1"
 
   spec.add_development_dependency "json", "~> 2.1"
   spec.add_development_dependency "kramdown", "~> 2.1"
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "cucumber", [">= 2.4", "< 6.0"]
   spec.add_runtime_dependency "rspec-expectations", "~> 3.4"
   spec.add_runtime_dependency "thor", "~> 1.0"
+  spec.add_runtime_dependency "bundler", "~> 2.1"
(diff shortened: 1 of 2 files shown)


luke-hill:
Only debate is whether it should be done this way or detected at runtime and required via another means?

mvz:
> Only debate is whether it should be done this way or detected at runtime and required via another means?

This is a tricky one. On the one hand, everyone uses Bundler. On the other hand, the error you encountered actually tells me not everyone uses Bundler. But then, why force it on people?

Also, I vaguely remember the transition from Bundler 1.x to 2.0 was difficult because many gems depended on Bundler 1.x.

Finally, if people want to use the Aruba::Api::Bundler they'll probably have Bundler installed.

So, I'm leaning toward doing some form of runtime check here.

luke-hill:
This error came about when running tests through RM. Which is a reasonably valid use case.

If we want to alter the runtime dependency to be more lax I'm happy to do that. But this does need adding in some way shape or form.

mvz:
@luke-hill what is RM?

luke-hill:
Sorry. RubyMine, it's a popular IDE for ruby. So for me running Aruba I just click on a feature and click "run"

mvz:
Ok, so using RubyMine leads to an error, but using just cucumber works?

luke-hill:
I believe so yes. But I only tested that adding this fixes the problem. I didn't dig deep into it as it felt not that big of an issue to add this as a runtime dep, because technically it is a runtime dep in this one area.

Can amend if you would prefer to something a bit more dynamic.

mvz:
Ok, so just to be clear:
- In a project using Aruba, running a scenario using `unset_bundler_env_vars` through RubyMine fails.
- If you add bundler as a dependency to that project, running the scenario through RubyMine no longer fails.
Correct?

aruba.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "cucumber", [">= 2.4", "< 6.0"]
   spec.add_runtime_dependency "rspec-expectations", "~> 3.4"
   spec.add_runtime_dependency "thor", "~> 1.0"
+  spec.add_runtime_dependency "bundler", "~> 2.1"
```
mvz, line 29: Let's allow both bundler 1.x and 2.x since 1.x may be in use on older Rubies. So something like `[">= 1.0", "< 3.0"]`.

Also rubocop wants the dependencies to be sorted alphabetically.

mvz (changes requested):
This makes sense from a dependency standpoint, whether we understand RubyMine's behaviour or not.

I have just one small nit; once that's fixed this can be merged.

mvz (approved):

PR #811, opened 2021-06-13
Allow testing with different versions of runtime dependencies

### LINKED ISSUE #729: Test with multiple versions of Cucumber

mvz:
## Summary

Add appraisal to easily allow testing with older versions of dependencies.

## Details

This introduces appraisals for defining specific versions of (runtime) dependencies for testing against. Normal development can still use the main Gemfile, but by using appraisal we can also test with other, older versions of the dependencies.

The GitHub Actions workflow file is also updated to test at least once with each defined appraisal.

Finally, the main Gemfile has been updated to not include any Ruby logic but instead rely on the platform setting to avoid installing byebug on JRuby.

## Motivation and Context

This fixes #729.

## How has this been tested

Succesfully ran `bundle exec appraisal` to install all gems, and `bundle exec appraisal cucumber_6 rake` to run tests.

## Types of changes

* Internal change (developer experience)

## Checklist


aurelien-reeves:
Great! Thank you @mvz

Issue #806, opened 2021-05-13
Release aruba/rspec as a standalone gem without cucumber


caalberts:
## Summary

Release `aruba-rspec` gem which contains only relevant code for `aruba/rspec`.

## Expected Behavior

When I need to use `aruba` with `RSpec`, I should be able to install the gem without cucumber.

Given the following Gemfile:

```rb
gem 'aruba-rspec'
```

When `bundle install`

Then `Gemfile.lock` should contain:
```
    aruba (1.1.0)
      childprocess (>= 2.0, < 5.0)
      contracts (~> 0.16.0)
      rspec-expectations (~> 3.4)
      thor (~> 1.0)
```

## Current Behavior

Given the following Gemfile:

```rb
gem 'aruba'
```

When `bundle install`

Then `Gemfile.lock` should contain:
```
    aruba (1.1.0)
      childprocess (>= 2.0, < 5.0)
      contracts (~> 0.16.0)
      cucumber (>= 2.4, < 7.0) # cucumber is also installed
      rspec-expectations (~> 3.4)
      thor (~> 1.0)
```

## Context & Motivation

When `aruba/rspec` is used, there doesn't seem to be a need for the `cucumber` gem to be installed as well. Releasing `aruba-rspec` on its own would reduce the bundle size.

mvz:
Please see #402 for previous discussion about splitting aruba.

The name `aruba-rspec` is taken, unfortunately, but apart from that, I'm all for this.

Pull requests welcome, but they'll have to include a migration path.

mvz:
Aruba's code currently only depends directly on `cucumber-core` so we could drop the full `cucumber` dependency at least.

Issue #402, opened 2016-05-10
Split up aruba into separate gems for "better" dependencies


dg-ratiodata:
## Summary

Split up aruba into separate gems: aruba-core, aruba-rspec, aruba-cucumber.
Ping @mattwynne 

## Current Situation

We now support integration into `cucumber`, `rspec` and `minitest`. At least for the first two ones we would normally define dependencies on "gem"-level. Today we only require `cucumber` to be installed. This is not required by users using `aruba` with `rspec` only.

~~~
~~~

## Possible Solution

1. Create a gem `aruba` with dependencies to:

  * aruba-cucumber
  * aruba-rspec
  * aruba-core
   
  This gem only includes the executable `aruba`. Otherwise bundler will complain or we rename it to `aruba-cli` and put it into a separate gem as well.

2. Create a gem `aruba-cucumber` with dependencies to:

  * aruba-core
  * cucumber

3. Create a gem `aruba-rspec` with dependencies to:

  * aruba-core
  * rspec

4. Create a gem `aruba-core` with the rest of the dependencies of the current `aruba`-gem.

5. Maybe: Create a gem `aruba-cli` which only contains cli-commands for `aruba`.

# Open Questions

* [ ] It makes sense to do this
* [ ] Shall we have separate repositories or place everything in a single repository

## Benefits

Users only have depencies installed they really need.

maxmeyer:
@mvz @junaruga @olleolleolle What do you think about this? This does not to be done till 1.0.0 can be also introduced later, 1.1.0 or even 2.1.0 (due to drop of rubies)

mvz:
In general, I think this makes sense.

I don't think there's a need to split of `aruba-rspec`, because Aruba already depends on only `rspec-expectations` and would need that in a hypothetical `aruba-core`. So, just splitting off `aruba-cucumber` should be enough. Since `aruba` would then not depend on `aruba-cucumber`, this would be a breaking change hence something for 2.0 (unless it is done right now).

I would recommend keeping everything in the same repository. This makes synchronizing development between the parts easier and preserves history. I've split a project apart and then grafted the repositories back together because `git blame` had become completely useless.

mvz:
With #806, there is no part of Aruba apart from the step definitions that depends on Cucumber.

mvz:
> With [#806](https://github.com/cucumber/aruba/issues/806), there is no part of Aruba apart from the step definitions that depends on Cucumber.

I probably meant #813 here.

[...]

PR #813, opened 2021-06-14
Use Ruby's built-in windows platform detection

mvz:
## Summary

Use Ruby's built-in windows platform detection.

## Details

Use `Gem.win_platform?` instead of piggy-backing on Cucumber to check if we're running on Windows.

## Motivation and Context

This avoids the need to load Cucumber when using Aruba with RSpec. A step toward fixing #402 and #806.

## How Has This Been Tested?

Mostly CI.

## Types of changes

- [ ] Documentation update

## Checklist:

### DIFF
--- a/lib/aruba/platforms/unix_platform.rb
+++ b/lib/aruba/platforms/unix_platform.rb
@@ -31,7 +31,7 @@ module Aruba
     # @private
     class UnixPlatform
       def self.match?
-        !Cucumber::WINDOWS
+        !Gem.win_platform?
       end
 
       def environment_variables
--- a/lib/aruba/platforms/windows_platform.rb
+++ b/lib/aruba/platforms/windows_platform.rb
@@ -1,5 +1,3 @@
-require "cucumber/platform"
-
 require "aruba/platforms/unix_platform"
 require "aruba/platforms/windows_command_string"
 require "aruba/platforms/windows_environment_variables"
@@ -20,7 +18,7 @@ module Aruba
     # @private
     class WindowsPlatform < UnixPlatform
       def self.match?
-        Cucumber::WINDOWS
+        Gem.win_platform?
       end
 
       # @see UnixPlatform#command_string
--- a/spec/aruba/api/commands_spec.rb
+++ b/spec/aruba/api/commands_spec.rb
@@ -46,10 +46,10 @@ RSpec.describe Aruba::Api::Commands do
     end
 
     context "when running a relative command" do
-      let(:cmd) { Cucumber::WINDOWS ? "bin/testcmd.bat" : "bin/testcmd" }
+      let(:cmd) { Gem.win_platform? ? "bin/testcmd.bat" : "bin/testcmd" }
 
       before do
-        if Cucumber::WINDOWS
+        if Gem.win_platform?
           @aruba.write_file cmd, <<~BAT
             exit 0
           BAT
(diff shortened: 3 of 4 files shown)


voxik:
Sadly, this interchange dependency on Cucumber for dependency on RubyGems. While I understand, that RubyGems is mostly available, I don't think it needs to be always the case, e.g. in case `bundle install --standalone` is used.

voxik:
I am thinking, would it be acceptable to use `RSpec::Support::OS.windows?` instead?

mvz:
Unfortunately, rspec-support is only for internal use by rspec, so not something we can rely on.

mvz:
`ruby -e 'puts Gem.win_platform?'` seems to run fine even on Ruby 2.5.0, so I don't think this is a problem. If you can produce a test case where `Gem` is not available, I can investigate and look into alternatives.

jackorp:
> `ruby -e 'puts Gem.win_platform?'` seems to run fine even on Ruby 2.5.0, so I don't think this is a problem. If you can produce a test case where `Gem` is not available, I can investigate and look into alternatives.

This is the issue for example with the rspec-core v3.10 spec suite, which uses the `--disable-gems` ruby option, resulting in Rubygems not being available.

When I follow steps for running the CI with unrestricted Aruba version (which pulls in Aruba 2.0), available in my branch: https://github.com/jackorp/rspec-core/tree/unrestrict-aruba
```
$ bundle install --standalone
$ bundle binstubs --all
$ script/run_build
```
then I get an exception regarding the Gem module:
```
$  script/run_build
============= Starting binstub check ===============
Checking required binstubs
============= Ending binstub check ===============
============= Starting specs ===============
/home/jprokop/projects/ruby/rspec-core/bin/rspec

An error occurred while loading spec_helper.
Failure/Error: require 'aruba/api'

NameError:
  uninitialized constant Aruba::Platforms::WindowsPlatform::Gem
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/platforms/windows_platform.rb:21:in `match?'
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/platform.rb:8:in `each'
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/platform.rb:8:in `find'
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/platform.rb:8:in `<module:Aruba>'
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/platform.rb:5:in `<top (required)>'
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/api.rb:6:in `require'
# ./bundle/ruby/3.0.0/gems/aruba-2.0.0/lib/aruba/api.rb:6:in `<top (required)>'
# ./spec/support/aruba_support.rb:28:in `require'
# ./spec/support/aruba_support.rb:28:in `block in <module:ArubaLoader>'
# ./spec/support/aruba_support.rb:27:in `<module:ArubaLoader>'
# ./spec/support/aruba_support.rb:25:in `<top (required)>'
# ./spec/spec_helper.rb:24:in `require'
# ./spec/spec_helper.rb:24:in `block in <top (required)>'
# ./spec/spec_helper.rb:17:in `map'
# ./spec/spec_helper.rb:17:in `<top (required)>'
# script/rspec_with_simplecov:42:in `load'
# script/rspec_with_simplecov:42:in `<main>'
No examples found.


Finished in 0.00008 seconds (files took 0.13783 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

mvz:
Thanks, @jackorp, I made an issue to track this.

PR #850, opened 2022-01-04
Test with Ruby 3.1 in CI

mvz:
## Summary

Test with Ruby 3.1 in CI

## Details

Add Ruby 3.1 to the CI matrix; also add `--disable-did-you-mean` option to avoid extra output that disturbs the cucumber scenario's.

## Motivation and Context

Aruba should work with Ruby 3.1.

## How Has This Been Tested?

CI

## Types of changes

- [ ] Documentation update

## Checklist:

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -14,6 +14,7 @@ on:
 
 env:
   CUCUMBER_PUBLISH_QUIET: true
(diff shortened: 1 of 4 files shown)


olleolleolle:
macos + windows could also get the 3.1 into its lists.

mvz:
The simplecov configuraton needs to be adjusted so we don't call `.start` more than once but still keep the different labels.

PR #851, opened 2022-01-04
Inherit default RuboCop exclusions

mvz:
## Summary

Simplify RubyCop config by inheriting default exclusions

## Details

RubyCop excludes tmp/ and vendor/ by default, so remove them from the local exclusions list and make sure they remain even if we add new local exclusions, by specifying the 'merge' inherit mode for exclusions.

## Motivation and Context

Simpler is better.

## How Has This Been Tested?

N/A

## Types of changes

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,12 +1,13 @@
+inherit_mode:
+  merge:
+    - Exclude
+
 require:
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
TIL! This is a great way of doing it!

PR #855, opened 2022-01-15
Use just IRB for debugging

mvz:
## Summary

Use plain irb for debugging and for demonstrating debugging

## Details, Motivation and Context

Development of pry has stalled and it is suggested that debug in combination with irb is the future. For now, we cannot use debug yet since only a very old version runs on Ruby 2.5.

For now, this change removes the pry, pry-doc, byebug and pry-byebug dependencies. If needed for local development, these can be added back via Gemfile.local.

Debugging inside a scenario using Aruba is now demonstrated using plain irb. This has the added advantage of also working on JRuby.

Once support for Ruby 2.5 has been dropped, the new debug gem can be added as a permanent dependency to provide extended debugging features by default.

## How Has This Been Tested?

I did a simple debug session with debug. This worked on Ruby 3.1, but not really on 2.5.

I ran the adjusted scenario on Ruby 3.1 and JRuby.

## Types of changes

- [x] Documentation update

## Checklist:

### DIFF
--- a/features/03_testing_frameworks/cucumber/steps/command/debug_your_command_in_aruba.feature
+++ b/features/03_testing_frameworks/cucumber/steps/command/debug_your_command_in_aruba.feature
@@ -7,7 +7,6 @@ Feature: Debug your command in cucumber-test-run
     Given I use a fixture named "cli-app"
     And the default aruba exit timeout is 60 seconds
 
-  @unsupported-on-platform-java
   Scenario: You can use a debug repl in your cli program
 
     If you want to debug an error, which only occurs in one of your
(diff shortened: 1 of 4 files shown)


features/03_testing_frameworks/cucumber/steps/command/debug_your_command_in_aruba.feature
```diff
@@ -53,11 +50,15 @@ Feature: Debug your command in cucumber-test-run
...
-    """
-    [1] pry(main)> foo
-    => "hello"
-    [2] pry(main)> exit
-    """
+      """
+       => 8: binding.irb
+          9: 
+          10: exit 0
+      Switch to inspect mode.
```
mvz, line 57: This looks weird.

mvz: There doesn't seem to be much we can do about this. Maybe we should keep `pry` around for now and just drop `byebug`.

mvz:
Pry was removed in #914 and #939 so this pull request is not needed anymore.

PR #858, opened 2022-01-16
Ensure Gem.win_platform? is available

### LINKED ISSUE #852: Using Gem.win_platform? may lead to issues
**Describe the bug**
When `--disable-gems` is used, `Gem` is not available.

**To Reproduce**
See https://github.com/cucumber/aruba/pull/813#issuecomment-1009883293

**Expected behavior**
Aruba works and correctly determines presence of windows.

mvz:
## Summary

Ensure `Gem.win_platform?` is available by explicitly requiring `rubygems`.

## Motivation and Context

Even though rubygems is nearly always loaded by ruby automatically, it is possible to avoid doing so. Therefore, to make platform detection work, we have to require "rubygems" explicitly.

Fixes #852.

## How Has This Been Tested?

Specs were added that exercise the relevant functionality for the case where rubygems is not loaded.

## Types of changes

## Checklist:

### DIFF
--- a/lib/aruba/platforms/unix_platform.rb
+++ b/lib/aruba/platforms/unix_platform.rb
@@ -1,4 +1,5 @@
 require "rbconfig"
+require "rubygems"
 require "pathname"
 
 require "aruba/aruba_path"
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Love the test!

PR #864, opened 2022-03-12
Loosen development dependency version constraints

mvz:


olleolleolle (approved):
Good choice. 

If this means some old Ruby versions are no longer supported, we shouldn't support those Ruby versions in our test suite.

mvz:
It seems we're fine for now when it comes to Ruby versions.

olleolleolle:
@mvz Green! Hooray!

PR #881, opened 2022-12-25
Use RSpec annotations instead of included class to trigger RSpec hooks

mvz:
## Summary

Use `type: :aruba` to trigger all RSpec hooks in the default integration, replacing the check for inclusion of Aruba::Api.

## Details

When using the default aruba/rspec integration, the initial inclusion of `Aruba::Api` was triggered by a `type: :aruba` annotation to the spec, but the other hooks were filtered by a check for inclusion of Aruba::Api.

Since we already use :type to trigger inclusion of Aruba::Api, we can safely use it to activate the around hooks as well, thus simplifying the code

This also increases safety for custom integrations, since having those work correctly depended on "aruba/rspec" never being loaded.

## Motivation and Context

This improves simplicity of the code by relying on existing mechanisms instead of introducing our own.

## How Has This Been Tested?

CI will check this

## Types of changes

- (Maybe) bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/lib/aruba/rspec.rb
+++ b/lib/aruba/rspec.rb
@@ -10,47 +10,35 @@ RSpec.configure do |config|
   config.include Aruba::Api, type: :aruba
 
   # Setup environment for aruba
-  config.around :each do |example|
-    if self.class.include? Aruba::Api
-      setup_aruba
+  config.around :each, type: :aruba do |example|
+    setup_aruba
 
-      # Modify PATH to include project/bin
-      prepend_environment_variable(
-        'PATH',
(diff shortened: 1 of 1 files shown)


mvz:
Needs work since the specs secretly relied on the hooks in aruba/rspec 😢

olleolleolle (approved):
🚀 Looks super!

(Well, the idea is super. Some order-of-operations may still be in our way.)

mvz:
Now that #1067 is merged, this should work properly.

PR #886, opened 2023-01-03
Dependabot to Renovate

mattwynne:
See https://github.com/cucumber/common/issues/2143


mattwynne:
> Hi @mattwynne what is the reason to prefer Renovate over Dependabot?

Just consistency. We're using Renovate everywhere else in the org now for all other languages, and we've figured out the hurdle for Ruby too. So for those of us working across multiple languages it just makes it easier to only work with one tool.

mvz:
Alright, that makes sense.

mvz:
Ah, apparently not, I just got the first Renovate pull request :+1:

PR #892, opened 2023-01-21
Replace ChildProcess with Process.spawn

### LINKED ISSUE #884: Use Process.spawn instead of the child_process gem
Since aruba is Ruby 2.6+ it can use Process.spawn which exists since 2.4.

Process.spawn covers the whole API of child_process, so there is no need for that dependency and it should be fairly easy to remove it.
child_process is unfortunately not very actively maintained, and that causes issues, for instance https://github.com/oracle/truffleruby/issues/1525 / https://github.com/enkessler/childprocess/issues/172.

### 🤔 What's the problem you're trying to solve?

Getting aruba and simplecov to work well under modern Rubies.

mvz:
## Summary

This replaces ChildProcess with Process.spawn for the purpose of starting, stopping and controlling the processes under test.

## Details

To make this change be the least invasive, I created a new class that emulates the interface provided by ChildProcess, with a few differences where this was more practical. For example, where ChildProcess provides an #io method that returns an object giving access to stdout etc., this new class simply privides #stdout and related methods directly.

This also removes the childprocess dependency.

## Motivation and Context

This fixes #884.

## How Has This Been Tested?

I ran the full test suite with MRI 3.2 and JRuby 9.4.

I also added specs for SpawnProcess so its tested more fully in the RSpec suite.

## Types of changes

## Checklist:

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -2,25 +2,20 @@
 
 appraise "cucumber_4" do
   gem "cucumber", "~> 4.0"
-  gem "childprocess", "~> 2.0.0"
 end
 
 appraise "cucumber_5" do
   gem "cucumber", "~> 5.0"
-  gem "childprocess", "~> 3.0.0"
 end
 
 appraise "cucumber_6" do
   gem "cucumber", "~> 6.0"
-  gem "childprocess", "~> 4.0"
 end
 
 appraise "cucumber_7" do
   gem "cucumber", "~> 7.0"
-  gem "childprocess", "~> 4.0"
 end
 
 appraise "cucumber_8" do
   gem "cucumber", "~> 8.0"
-  gem "childprocess", "~> 4.0"
 end
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -22,7 +22,6 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_runtime_dependency "bundler", [">= 1.17", "< 3.0"]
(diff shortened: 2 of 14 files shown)


olleolleolle:
Many high-fives for this amazing piece!

olleolleolle:
@mvz This is quite exciting - what could be the signal that triggers the EINVAL? Would you care to pair while finding out?

olleolleolle:
@mvz  Around the signals: The Thin web server also skips handling HUP on Windows.

https://github.com/macournoyer/thin/blob/02f0a3a72620313c69a1d797671cc522399d9b4c/lib/thin/server.rb#L241

mvz:
I made it log the signal and it's the TERM signal which I've currently mapped to signal 3 for Windows (I'm going to undo that mapping since just sending the signal name should work just fine.

However, now I found this explanation of how things work on Windows: https://github.com/enkessler/childprocess/issues/114#issuecomment-293898550. So, I'm going to try that experiment with that now.

I unfortunately have no Windows machine available so debugging is via GitHub Actions.

I'll may have to skip the spec that tests HUP on Windows since the whole point is to send a custom signal, and Windows seems to only support KILL.

olleolleolle:
@mvz Great job!

Also, interesting to learn from enkessler here.

mvz:
Thanks @olleolleolle. pevtsoff does not seem very active in the Ruby world at the moment. I've asked the submitter of #884 to comment, though.

PR #896, opened 2023-02-12
Fix new RuboCop offenses

mvz:
## Summary

Fix some new RuboCop offenses

## Details

- Autocorrect Performance/StringInclude
- Format simple table entirely using the format string
- Exclude fixtures from DevelopmentDependencies cop
- Regenerate RuboCop to-do file

## Motivation and Context

The RuboCop task was failing due to updates in RuboCop.

## How Has This Been Tested?

I ran RuboCop

## Types of changes


bin/cucumber
```diff
@@ -15,7 +15,7 @@ ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",
 bundle_binstub = File.expand_path("bundle", __dir__)
 
 if File.file?(bundle_binstub)
-  if /This file was generated by Bundler/.match?(File.read(bundle_binstub, 300))
+  if File.read(bundle_binstub, 300).include?("This file was generated by Bundler")
```
olleolleolle, line 18: Perhaps we don't want to lint the binstub files generated by other gems?

mvz: Yes, that's a good point.

PR #897, opened 2023-02-25
Remove yard-junk and pry-doc dependencies

mvz:
## Summary

Removes dependencies that pull in yard and hence webrick

## Details

By removing yard-junk and pry-doc, we drop the dependency on YARD. This is done mainly to avoid the outdated dependency on webrick that YARD pulls in in turn.

Running yard-junk was not part of CI, so we lose nothing there. When desired, yard junk can be checked for by running `yard doc --plugin junk`.  Documentation provided by pry-doc can easily be read outside of the REPL.

## Motivation and Context

Dropping an outdated dependency.

## How Has This Been Tested?

N/A

## Types of changes

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,7 +8,6 @@ Bundler.setup
 require "cucumber/rake/task"
 require "rspec/core/rake_task"
 require "rubocop/rake_task"
-require "yard-junk/rake"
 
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Good idea!

PR #899, opened 2023-04-28
Simplify use of match_array to just eq

mvz:
## Summary

Replace `match_array` with just `eq`.

## Details

Both elements of the array are the same, so testing for equality is just fine.

## Motivation and Context

This fixes a RuboCop RSpec/MatchArray offense.

## How Has This Been Tested?

I ran the relevant spec file.

## Types of changes


olleolleolle (approved):
🧹 Keeping tidy!

Issue #910, opened 2023-09-24
cucumber scenarios fail with ruby 3.3


mtasaka:
With ruby 3.3 ( I tried https://github.com/ruby/ruby/commit/2ceb5363f9b15cffd3901220463a1175a32b89f7 ) , aruba ( https://github.com/cucumber/aruba/commit/e9192bec97e8d26c7382efee5f8bbd06fb96a924 ) cucumber test fails like:

```
Failing Scenarios:
cucumber features/03_testing_frameworks/cucumber/steps/command/debug_your_command_in_aruba.feature:11 # Scenario: You can use a debug repl in your cli program
cucumber features/03_testing_frameworks/cucumber/steps/command/wait_for_output_of_command.feature:24 # Scenario: Detect subset of multiline output
cucumber features/03_testing_frameworks/cucumber/steps/command/wait_for_output_of_command.feature:47 # Scenario: Stop when output appears
cucumber features/06_use_aruba_cli/open_console.feature:6 # Scenario: Start console
cucumber features/06_use_aruba_cli/open_console.feature:14 # Scenario: Show help
cucumber features/06_use_aruba_cli/open_console.feature:31 # Scenario: Show methods
cucumber features/06_use_aruba_cli/open_console.feature:44 # Scenario: Has its own history file
```

I think 4 `open_console.feature` scenarios failures are due to https://github.com/ruby/irb/pull/613 , for other 3 failures I have no idea for now.

Details are:
```
expected "Feature: Exit status in debug environment\n\n  @debug\n  Scenario: Run program with debug code # fea...                │\n└──────────────────────────────────────────────────────────────────────────────┘" to string includes: "[1] pry(main)> foo\n=> \"hello\"\n[2] pry(main)> exit"
Diff:
@@ -1,4 +1,37 @@
-[1] pry(main)> foo
-=> "hello"
-[2] pry(main)> exit
+Feature: Exit status in debug environment
+
+  @debug
+  Scenario: Run program with debug code # features/debug.feature:4
+
+From: /builddir/build/GIT/aruba/tmp/aruba/cli-app/bin/aruba-test-cli:9 :
+
+     4: $stdout.sync = true
+     5: 
+     6: foo = 'hello'
+     7: 
+     8: require 'pry'
+ =>  9: binding.pry
+    10: 
+    11: exit 0
+
[...]

PR #914, opened 2023-12-17
Prepare for Ruby 3.3

mvz:
## Summary

Prepare for upcoming Ruby 3.3

## Details

This pull request consists of the following changes:

- Test with Ruby 3.3.0-preview3 in CI. This tests that Aruba is ready for Ruby 3.3
- Use irb instead of pry in the interactive debugging scenario. This fixes an issue in the Cucumber scenario's. This issue occurs because our output streams are not reall TTYs. See #910 for details.
-  Stop requiring 'irb/ext/save-history' extension in Aruba console. This extension has been renamed and did not do what its old name said.
- Ensure irb uses a consistent prompt format in tests on all Ruby versions.

## Motivation and Context

Mainly, #910.

## How Has This Been Tested?

I ran the Cucumber scenario with the added gem in Ruby 3.3.0-preview3. I also tried just setting TERM=dumb as in https://github.com/rails/rails/pull/48369 but that didn't work.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -23,7 +23,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby: ["3.1", "3.2", "jruby-9.4"]
+        ruby: ["3.1", "3.2", "jruby-9.4", "3.3.0-preview3"]
         appraisal: [cucumber_8, cucumber_9]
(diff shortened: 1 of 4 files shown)


mvz:
Ugh ...

PR #921, opened 2024-04-01
Fix minitest setup code and make minitest support scenarios run at least one test

mvz:
## Summary

Make minitest scenarios run at least one test so test summary is displayed.

## Details

This change fixes the name of the test method in the minitest scenario so minitest actually recognizes it as a test. It also updates the setup code which was outdated. This was never noticed because it was not run.

It also updates the minitest initializer so it generates a dummy test, and fixes the test setup that it generates.

## Motivation and Context

Due to how test success is checked, the failure to run at least one test was never noticed. A change in behavior in minitest exposed the problem by making nearly all CI jobs fail. See https://github.com/minitest/minitest/pull/986.

## How Has This Been Tested?

The failing scenarios now pass.

## Types of changes

- [ ] Documentation update

## Checklist:

### DIFF
--- a/features/01_getting_started_with_aruba/supported_testing_frameworks.feature
+++ b/features/01_getting_started_with_aruba/supported_testing_frameworks.feature
@@ -72,15 +72,15 @@ Feature: Supported Testing Frameworks
       include Aruba::Api
 
       def setup
-        aruba_setup
+        setup_aruba
       end
 
-      def getting_started_with_aruba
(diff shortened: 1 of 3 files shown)


mvz:
Another scenario is failing which really should have zero tests run :cry:

mvz:
The RuboCop failure will be fixed in a separate pull request.

PR #926, opened 2024-04-01
Unlink Tempfiles after use in SpawnProcess

mvz:
## Summary

Explictly call `#unlink` on Tempfiles used in SpawnProcess, instead of waiting for the finalizer.

## Details

Adds a call to `#unlink` right after reading the Tempfile and storing the contents in a variable.

## Motivation and Context

On Windows, the finalizer for Tempfile will raise an error if Errno::ENOACCES is raised when removing the file. However, if this happens when #unlink is called, the error is caught.

This change calls `#unlink` on the Tempfiles that are used to store the output of spawned processes. This means the finalizer will not need to unlink the file, so any Errno::ENOACCES errors during unlinking will be caught.

The raised errors did not make any specs fail, but they did appear in the rspec output.

## How Has This Been Tested?

I ran the specs. In CI, we will see if the backtraces are gone.

## Types of changes

- [ ] Documentation update

## Checklist:

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -17,7 +17,7 @@ Metrics/BlockLength:
 
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 145
+  Max: 146
 
 # Configuration parameters: AllowedMethods, AllowedPatterns.
 Metrics/CyclomaticComplexity:
@@ -101,8 +101,6 @@ RSpec/InstanceVariable:
     - 'spec/aruba/api/filesystem_spec.rb'
     - 'spec/aruba/api_spec.rb'
     - 'spec/aruba/event_bus_spec.rb'
-    - 'spec/aruba/matchers/command/have_output_size_spec.rb'
-    - 'spec/aruba/matchers/command_spec.rb'
     - 'spec/aruba/matchers/directory_spec.rb'
     - 'spec/aruba/matchers/file_spec.rb'
     - 'spec/aruba/matchers/path_spec.rb'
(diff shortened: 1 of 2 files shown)


mvz:
Good theory but this doesn't help 😞

mvz:
Here's the `#unlink` method, and it is now clear why this PR doesn't help:

```ruby
  def unlink
    return if @unlinked
    begin
      File.unlink(@tmpfile.path)
    rescue Errno::ENOENT
    rescue Errno::EACCES
      # may not be able to unlink on Windows; just ignore
      return
    end
    ObjectSpace.undefine_finalizer(self)
    @unlinked = true
  end
```

If unlink fails due to `Errno::EACCES`, the finalizer is not removed, and that finalizer will try again to unlink the Tempfile. As a result, `Errno::EACCES` errors will still occur there and cause backtraces to be printed.

This method also has the following comment:

```ruby
  # However, unlink-before-close may not be supported on non-POSIX operating
  # systems. Microsoft Windows is the most notable case: unlinking a non-closed
  # file will result in an error, which this method will silently ignore. If
  # you want to practice unlink-before-close whenever possible, then you should
  # write code like this:
```

So the `Errno::EACCES` error is expected to come from the Tempfile still being open.

olleolleolle:
(Good investigation!)

mvz:
I have looked further at the documentation for Tempfile, and the since Ruby 3.4, `Tempfile.create` is recommended over `Tempfile.new`, since the former creates a `File` object, without unlinking at garbage collection time, and without the overhead of delegation. For the solution used in this pull request, it means that, if unlinking fails right after reading the temporary file's content, it will not be tried again, and no backtraces will be printed.

In addition, `Tempfile.create` supports an option `anonymous` that unlinks the file before returning it, avoiding the need for explicit unlinking entirely. This option also works on Windows.

mvz:
Using `Tempfile.create` worked. See #1059.

PR #935, opened 2024-07-10
Refactor: Switch out ~ expression in tag for not statement

luke-hill:
## Summary

Switch out legacy tag expression syntax - This is needed to fix up some tech debt in cucumber (Which sanitizes it)

## Details

As written above

## Types of changes

- [ ] Documentation update

## Checklist:


luke-hill:
Hi @mvz 
I'm not sure what's going on here with the CI e.t.c. so I've merged this as it's a tiny bugfix. I'll let you decide if anything needs remedying after the fact as I don't quite know where the CI is in this repo (I notice lots of the PR's are :x: too)

mvz:
@luke-hill one thing that makes the PR fail is the change in quote characters in `lib/aruba/cucumber/hooks.rb`. Did that happen automatically due to some some editor setting?

Perhaps I need to adjust the RuboCop config so it matches up with the rest of Cucumber. Suggestions welcome!

luke-hill:
In terms of styling. We use single quotes in all ruby projects. Also we don't adhere to the gemspec rule about dev dependencies in the gemfile (No idea why that's a default).

I can amend the quotes. I think I clicked something in my editor for that, as I didn't use rubocop on the CLI. The other one I've no idea on though

mvz:
@luke-hill allright, here's my plan:
- I fix the complaint from rubocop about add_runtime_dependency
- I update Aruba's rubocop config to be (more) in line with the ruby cucumber gems and autocorrect the code to match it
- You rebase this pull request, and the rubocop complaints should go away
- I merge the pull request

mvz:
Hi @luke-hill, I have finally done the second step in my plan above. Could you please rebase your branch on current main?

luke-hill:
Will do.

It's also made me think we should have an overall cucumber style for rubocop as a separate gem that way it saves the hassle of setting things up each time. But I'm not sure when I'll get around to it.

luke-hill:
Yep, back now so will get this fixed soon

mvz:
Hi @luke-hill, the merge commit seems to contain surprising and unrelated changes making it hard to review this. Can you please rebase the original two commits and make a new PR out of the change to the CI configuration?

luke-hill:
The merge commit is just a standard one that fixes the flow. I'm not sure why the CI flow is stuck. I'll get that pushed in once I have time this week

mvz:
@luke-hill oh, you're right. I shouldn't look at the merge commit on its own because it shows the changes from `main`.

mvz:
The reason it's stuck is due to the renaming of `checks` to `static_checks`.

mvz:
I would really like to merge this, so can you please just rebase the original two commits?

Issue #678, opened 2019-12-28
Drop support for separate console history file


mvz:
## Summary

Aruba console uses its own history file. I don't think it should.

## Expected Behavior

History for aruba console should be appended to the general IRB history.

## Current Behavior

Aruba console stores its history in `~/.aruba_history`.

## Possible Solution

Remove the code setting the relevant IRB options, as well as Aruba's configuration option.

## Context & Motivation

 Having this history separate makes it harder to re-use existing history from general IRB use. Other special-purpose consoles (e.g., `rails console`) also do not keep a separate history.

mvz:
We may even drop the whole `aruba console` feature, _and_ the aruba CLI.

PR #969, opened 2025-12-24
Make aruba console work on Ruby 4.0

mvz:
## Summary

Add explict irb dependency.

## Details

In Ruby 4.0, irb is no longer part of Ruby's default gems, so it must be added as an explicit dependency. This pull request adds the dependency, and also adds Ruby 4.0's latest preview release to the CI matrix.

## Motivation and Context

Ruby 4.0 is near!

## How Has This Been Tested?

I ran the test suite on Ruby 4.0.0-preview3.

## Types of changes

### DIFF
--- a/aruba.gemspec
+++ b/aruba.gemspec
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'bundler', '>= 1.17'
   spec.add_dependency 'contracts', ['>= 0.16.0', '< 0.18.0']
   spec.add_dependency 'cucumber', '>= 8.0', '< 11.0'
+  spec.add_dependency 'irb', '~> 1.16'
   spec.add_dependency 'rspec-expectations', '>= 3.4', '< 5.0'
   spec.add_dependency 'thor', '~> 1.0'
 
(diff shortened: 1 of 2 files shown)


aruba.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'bundler', '>= 1.17'
   spec.add_dependency 'contracts', ['>= 0.16.0', '< 0.18.0']
   spec.add_dependency 'cucumber', '>= 8.0', '< 11.0'
+  spec.add_dependency 'irb', '~> 1.16'
```
pirj, line 29: @mvz do you think this can be made `add_development_dependency`?

pirj: I see lib/aruba/console.rb, so probably not. Sorry to bother.

mvz: Indeed, this is unfortunately needed by the aruba console feature. However, I'm not really sure this feature is pulling its weight. Users can simply use `irb` themselves and use `irb`'s features to load and explore Aruba.

See also https://github.com/cucumber/aruba/issues/678#issuecomment-932762036

PR #1023, opened 2026-05-26
Add OpenSSF Scorecard badge

mpkorstanje:
### 🤔 What's changed?

* Added OpenSSF Scorecard badge
* Spruced up the header a bit
* Because the GitHub actions badge has a different style I've put it on a separate line.

### ⚡️ What's your motivation? 

We improved the Cucumber orgs security stance a bit. The OpenSSF Scorecard provides a way to display that stance as a number. 

Contribute to: https://github.com/cucumber/common/issues/2312

### 🏷️ What kind of change is this?
- :book: Documentation (improvements without changing code)

### ♻️ Anything particular you want feedback on?

@mvz up to you if you want to display it. 

Overall we also get low scores for code-review and not requiring PR's for all changes. This primarily caused by using Renovate to automatically update our dependencies. Personally, I don't think the requirements the scorecard puts down to get good marks here are worth it though for a mostly unpaid opensource project. But if you want to make improvements here you are of-course free to do so.

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,9 @@
-<h1 align="center">
-  <img src="https://raw.githubusercontent.com/cucumber/cucumber-js/46a5a78107be27e99c6e044c69b6e8f885ce456c/docs/images/logo.svg" alt="Cucumber logo" width="75">
-  <br>
-  Aruba
-</h1>
-<p align="center">
-
-**Test command-line applications with Cucumber-Ruby, RSpec or Minitest**
-
-</p>
-
-<div align="center">
+# Aruba
 
 [![MIT license](https://img.shields.io/badge/license-MIT-blue.svg)](https://raw.githubusercontent.com/cucumber/aruba/main/LICENSE)
 [![Gem Version](https://badge.fury.io/rb/aruba.svg)](http://badge.fury.io/rb/aruba)
 [![Support](https://img.shields.io/badge/cucumber-support-orange.svg)](https://cucumber.io/support)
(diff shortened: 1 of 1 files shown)


mvz:
Some thoughts:

This score embodies some useful rules so I'm open to displaying it. The score for Aruba is not that bad (7.1).

The UX of the score card is not great:

1. The `CRITICAL` label for the first item is much more visible than the `10` that says Aruba is fine (https://github.com/ossf/scorecard-webapp/issues/622, open since April 2024)
2. The sort-by drop down does not keep track of the current sorting and because of this you can't actually sort by name a-z.
3. You can't sort by score, making it harder to find actionable items
4. There's no link to OpenSSF, so visitors can be a bit lost about what this even is

I'd prefer not to have HTML in the `README.md`. My markdown linter complains about it and it makes the README harder to read in a text editor.

README.md
```diff
@@ -1,9 +1,24 @@
-# Aruba
+<h1 align="center">
+  <img src="https://raw.githubusercontent.com/cucumber/cucumber-js/46a5a78107be27e99c6e044c69b6e8f885ce456c/docs/images/logo.svg" alt="Cucumber logo" width="75">
+  <br>
+  Aruba
+</h1>
+<p align="center">
+
+**Test command-line applications with Cucumber-Ruby, RSpec or Minitest**
```
mvz: This line is not center-aligned for some reason.

mvz:
@mpkorstanje can you separate out the scorecard badge change from the fancy header bit? I'll merge the former but the latter gives too many problems with the tooling I use.

mvz:
Thanks, @mpkorstanje!

PR #1048, opened 2026-07-12
Stop any commands running after each spec in Aruba's spec suite

mvz:
## Summary

Clean up running commands after each spec.

## Details

On windows, commands left running prevent clean-up of tmp files used to store stdout and stderr. This change should ensure commands are stopped and tmp files closed before they're garbage-collected.

## Motivation and Context

The specs on windows report a lot of warnings about access being denied on tmp files. This should clean that up.

## How Has This Been Tested?

I ran one spec locally. This should mainly be tested by checking the logs in CI.

## Types of changes

## Checklist:

N/A

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --order random
 --warnings
+--format documentation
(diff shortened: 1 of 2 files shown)


mvz:
This does not have the desired effect.

mvz:
The tempfile issue has been fixed in #1059.

PR #992, opened 2026-02-14
Fix Errno::EACCES errors in tempfile finalizer on Windows

mvz:
## Summary

TBD

## Details

TBD

## Motivation and Context

The build logs on Windows are filled with errors raised the finalizers of the tempfiles we use to capture the output of processes in `SpawnProcess`.

This may be a genuine problem preventing Aruba from working properly on Windows: It possibly causes or is related to the Aruba scratch directory not getting cleaned up.

## How Has This Been Tested?

Unfortunately, we once again have to test this in CI.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Internal change (refactoring, test improvements, developer experience or update of dependencies)

## Checklist:

- [ ] I've added tests for my code
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1059, opened 2026-07-18
Use anonymous temporary files that are unlinked upon creation

mvz:
## Summary

Use anonymous temporary files used to capture STDOUT and STDERR.

## Details

First, in `SpawnProcess`, use `Tempfile.create` to create a temporary file that is not unlinked on garbage collection. This means we have control over the moment of unlinking. Then, to avoid the need to unlink entirely, use `Tempfile.create`'s new `anonymous` flag to create already unlinked temporary files safely on both Windows and Unixes.

Also, clean up use of `Tempfile` in an RSpec helper.

- **Use default rspec output capturing instead of custom implementation**
- **Use Tempfile.create so we gain control over unlinking**
- **Use Tempfile's anonymous feature that safely unlinks upon creation**

## Motivation and Context

This is a reprise of #926, however with the new insight that `Tempfile.create` creates a regular `File` object that does not auto-unlink upon garbage collection. This means we can attempt to unlink it right after use, and not worry about backtraces being generated at random later moments. With the new `anonymous` flag, we don't even need to do the unlinking ourselves. See also #992 for another attempt using `File.unlink`.

## How Has This Been Tested?

I ran the specs on older and newer rubies, including JRuby. I also checked the `/tmp` directory and with the `anonymous` option and sufficiently recent `tempfile`, no files are visible there during the run.

## Types of changes

## Checklist:

N/A


mvz:
Inspecting the Windows test log, I can now see that the stack traces due to the Tempfile unlink errors at garbage collection time are now gone :tada:. Merging!

PR #1073, opened 2026-08-02
Improve Aruba::Api#which and allow empty path

mvz:
## Summary

The implementations of `Aruba.platform.which` would raise an early exception of the path was empty. This is not really necessary, so this PR removes that limitation.

In addition, it makes `Aruba::Api#which` find relative paths from the working directory rather than from the root directory, making it consistent with `#run_command`.

## Details

For the main change, this just removes the check on empty path from the platform-specific implementations of `which`.

For the API change, this wraps the call to `Aruba.platform.which` in `#in_current_directory`, so relative paths are resolved with respect to the workspace.

- **Organize specs for Aruba::Api::Commands by method**
- **Make Aruba::Api#which look up relative commands in the workspace**
- **Allow path argument for which to be empty**

## Motivation and Context

Some specs were sometimes failing in CI on Windows due to PATH being empty.

## How Has This Been Tested?

I added specs for `which` and ran them.

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/lib/aruba/api/commands.rb
+++ b/lib/aruba/api/commands.rb
@@ -13,20 +13,26 @@ module Aruba
   module Api
     # Command module
     module Commands
-      # Resolve path for command using the PATH-environment variable
+      # Resolve path for command.
+      # - An absolute path resolves to the path itself
+      # - A relative path is resolved with respect to the current Aruba working
+      #   directory
+      # - Other paths are resolved using the path
       #
       # @param [#to_s] program
       #   The name of the program which should be resolved
       #
       # @param [String] path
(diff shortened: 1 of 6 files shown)


mvz:
`#which` is not currently documented but should be.

mvz:
Haha now the Windows build still fails due to path being nil but the error is different. The dilemma now is:

- Do I just call `#to_s` on the path to make the problem with `nil` go away?
- Or do I tediously investigate why the path is `nil`?

